PR #2, opened 2023-04-06
Ci improvements

mbouron:


configure.py
```diff
@@ -224,6 +232,19 @@ def _pkgconf_install(cfg):
...
+def _ffmpeg_install(cfg):
+    dirs = (
+        ("bin", "Scripts"),
+        ("lib", "Lib"),
+        ("include", "Include"),
+    )
+    cmds = []
+    for d in dirs:
+        src = op.join(cfg.externals[_FFMPEG_ID], d[0], "*")
+        dst = op.join(cfg.prefix, d[1])
```
ubitux, line 267: You can give meaning to each side of the tuples:
```python
for (src, dst) in dirs:
    src = op.join(cfg.externals[_FFMPEG_ID], src, "*")
    dst = op.join(cfg.prefix, dst)
```

configure.py
```diff
@@ -68,6 +68,12 @@
         dst_file="opengl-registry-d0342a4.zip",
         sha256="8e6efa8d9ec5ef1b2c735b2fc2afdfed210a3a5aa01c48e572326914e27bb221",
     ),
+    sdl2=dict(
+        version="2.26.5",
+        url="https://github.com/libsdl-org/SDL/releases/download/release-2.26.5/SDL2-devel-2.26.5-VC.zip",
```
ubitux, line 74: Isn't the `@VERSION@` accessible in `url` as well?

configure.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 #
 # Copyright 2021-2022 GoPro Inc.
+# Copyright Nope Project
```
ubitux, line 4: 2023

configure.py
```diff
@@ -49,6 +49,18 @@
...
+        url="https://github.com/KhronosGroup/EGL-Registry/archive/@VERSION@.zip",
+        dst_file="egl-registry-@VERSION@.zip",
+        sha256="b981b5a250d4f10284c670ae89f4a43c9eac42adca65306e5d56b921f609c46b",
+    ),
+    opengl_registry=dict(
+        version="d0342a4",
+        url="https://github.com/KhronosGroup/OpenGL-Registry/archive/@VERSION@.zip",
+        dst_file="opengl-registry-@VERSION@.zip",
+        sha256="8e6efa8d9ec5ef1b2c735b2fc2afdfed210a3a5aa01c48e572326914e27bb221",
+    ),
```
ubitux, line 71: I believe these need to be present in the clean rules like the other packages?

doc/usr/howto/installation.md
```diff
@@ -47,7 +47,7 @@ building and running the complete `nope.gl` stack.
   `C:\vcpkg`, then from Windows PowerShell:
     ```shell
     .\bootstrap-vcpkg.bat
-    .\vcpkg.exe install --triplet x64-windows opengl-registry ffmpeg[ffmpeg,ffprobe,zlib] sdl2 glslang
+    .\vcpkg.exe install --triplet x64-windows ffmpeg[ffmpeg,ffprobe,zlib] sdl2 glslang
```
ubitux: We need a few words to explain to the user what they're expected to do then

configure.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 #
 # Copyright 2021-2022 GoPro Inc.
+# Copyright 2023 Nope Project
```
ubitux, line 4: Usually the most up-to-date goes on top (see for example in `android_imagereader.c`)

doc/usr/howto/installation.md
```diff
@@ -47,11 +47,8 @@ building and running the complete `nope.gl` stack.
   `C:\vcpkg`, then from Windows PowerShell:
     ```shell
     .\bootstrap-vcpkg.bat
-    .\vcpkg.exe install --triplet x64-windows ffmpeg[ffmpeg,ffprobe,zlib] sdl2 glslang
+    .\vcpkg.exe install --triplet x64-windows sdl2 glslang
     ```
-- Add `C:\vcpkg\installed\x64-windows\tools\ffmpeg` path to your windows system
-  `%PATH%` environment variable (`ffmpeg` and `ffprobe` binaries must be
-  available in order to run the tests)
```
ubitux, line 54: Same as previously: a few words (and the cleanup rule in the configure)

configure.py
```diff
@@ -43,7 +43,14 @@
 _ROOTDIR = op.abspath(op.dirname(__file__))
 _SYSTEM = "MinGW" if sysconfig.get_platform().startswith("mingw") else platform.system()
 _RENDERDOC_ID = f"renderdoc_{_SYSTEM}"
+_FFMPEG_ID = f"ffmpeg_{_SYSTEM}"
```
ubitux, line 46: I don't think we will need FFmpeg manually pulled anywhere else; you can probably just use `ffmpeg` identifier, just like egl/opengl registry and pkgconf.

ubitux (changes requested):

configure.py
mbouron: Currently there is no clean rule for renderdoc. Creating one for every package deploying binaries files means tracking the files we copy ? Do we want that ?

doc/usr/howto/installation.md
mbouron: About what ? opengl-registry is pulled by `configure.py` automatically. The user does not need to do anything about it.

configure.py
mbouron: Done.

doc/usr/howto/installation.md
ubitux: Oh yeah no right, I'm an idiot, disregard that

configure.py
ubitux: I had the misconception that we were compiling something, which is not the case, so yeah we probably don't need a cleanup rule

configure.py
mbouron: Done.

ubitux (approved):
LGTM

PR #3, opened 2023-04-06
Scopes

ubitux:


libnopegl/src/pgcraft.c
```diff
@@ -556,7 +556,7 @@ static int inject_block(struct pgcraft *s, struct bstr *b,
     for (int i = 0; i < ngli_darray_count(&block->fields); i++) {
         const struct block_field *fi = &field_info[i];
         const char *type = get_glsl_type(fi->type);
-        if (named_block->variadic && fi->count && i == ngli_darray_count(&block->fields))
+        if (fi->count == NGLI_BLOCK_VARIADIC_COUNT)
```
mbouron, line 559: I think parts of this change belongs in the previous commit (adding NGLI_BLOCK_VARIADIC_COUNT). In this commit, you should only drop `named_block->variadic && fi->count && i == ngli_darray_count(&block->fields)` condition.

libnopegl/src/node_colorstats.c
```diff
@@ -0,0 +1,414 @@
...
+    const struct colorstats_opts *o = node->opts;
+    struct gpu_ctx *gpu_ctx = ctx->gpu_ctx;
+
+    /*
+     * Define workgroup size using limits. We pick a value multiple of the
+     * depth on purpose.
+     *
+     * GL_MAX_COMPUTE_WORK_GROUP_SIZE must be at least [128,128,64] (OpenGLES
+     * 3.1), but GL_MAX_COMPUTE_WORK_GROUP_INVOCATIONS (x*y*z) is only 128.
+     * Meaning that if we pick a work group size X=128, we will have to use Y=1
```
mbouron, line 196: Can you use `gpu_limits.max_compute_work_group_size` and `gpu_limits.max_compute_work_group_invocations` while still mentioning the OpenGLES constrains.

mbouron: Please also mention the Vulkan contraints which are the same as OpenGLES 3.1.

libnopegl/src/node_colorstats.c
```diff
@@ -0,0 +1,414 @@
...
+     * and Z=1.
+     */
+    const struct gpu_limits *limits = &gpu_ctx->limits;
+    const int max_group_size_x = limits->max_compute_work_group_size[0];
+    ngli_assert(max_group_size_x >= 128);
```
mbouron, line 201: Is this check really useful ? Is it to prevent bad drivers from running the code ? If so why not check the other limits ?

libnopegl/src/node_colorstats.c
```diff
@@ -0,0 +1,414 @@
...
+
+    return 0;
+}
+
+static int init_block(struct colorstats_priv *s, struct gpu_ctx *gpu_ctx)
+{
+    s->blk.data = NULL; /* We do not have any CPU data */
+
+    /*
+     * std430 layout is needed because the buffer has to be writable, and
```
mbouron, line 240: std430 has nothing to do with the buffer being writable nor the variadic declaration. You can do that with std140.

libnopegl/src/block.c
```diff
@@ -152,6 +160,14 @@ int ngli_block_add_field(struct block *s, const char *name, int type, int count)
 {
     ngli_assert(s->layout != NGLI_BLOCK_LAYOUT_UNKNOWN);
 
+    if (count == NGLI_BLOCK_VARIADIC_COUNT)
+        ngli_assert(s->layout == NGLI_BLOCK_LAYOUT_STD430);
```
mbouron, line 182: You can have variadic declarations with the STD140 layout.

libnopegl/src/node_colorstats.c
```diff
@@ -0,0 +1,414 @@
...
+static int alloc_block_buffer(struct ngl_node *node, int length)
+{
+    struct colorstats_priv *s = node->priv_data;
+
+    /* We assume a 8-bit sampling all the time for now */
+    s->depth = 1 << 8;
+
+    /*
+     * Horizontal length, minus 1 to reduce operations in the shader
+     * TODO add a vertical mode using the image height instead
```
mbouron, line 304: TODO`:` ?

libnopegl/src/node_renderother.c
```diff
@@ -104,6 +104,7 @@ struct render_common {
     int topology;
     const struct geometry *geometry;
     struct darray pipeline_descs;
+    struct ngl_node *extra_draw_node;
```
mbouron, line 111: Can we have the same mechanism as in `pass.c` ?

libnopegl/src/node_colorstats.c
ubitux: It's a bit redundant with the comment, but since the line below is `s->group_size = max_group_size_x >= 256 ? 256 : 128;` I wanted to show the developer that 128 is an always safe fallback value. Do you want me to remove it?

libnopegl/src/node_renderother.c
ubitux: You mean you'd prefer to have a `darray` here as well?

mbouron: Yup + the automatic resources registrations.

libnopegl/src/node_colorstats.c
mbouron: I'm in favor of removing it.

libnopegl/src/pgcraft.c
ubitux: Mmh yeah sure makes sense, history adjusted.

libnopegl/src/node_colorstats.c
ubitux: Done. Comment adjusted to:

```c
    /*
     * Define workgroup size using limits. We pick a value multiple of the
     * depth on purpose.
     *
     * The OpenGLES 3.1 and Vulkan core specifications mandate that
     * gpu_limits.max_compute_work_group_size must be least [128,128,64], but
     * gpu_limits.max_compute_work_group_invocations (x*y*z) minimum is only
     * 128. Meaning that if we pick a work group size X=128, we will have to
     * use Y=1 and Z=1. 128 remains an always safe value so we use that as
     * a fallback.
     */
```

libnopegl/src/node_colorstats.c
ubitux: Removed, though for some reason it seems that using std140 here is leading to errors such as `Assert buffer->usage & NGLI_BUFFER_USAGE_UNIFORM_BUFFER_BIT @ src/backends/gl/pipeline_gl.c:714`. For some reason it appears the pipeline wants to bind it as an UBO when the layout is std140. I couldn't remember nor find where we have this heuristic.

libnopegl/src/block.c
ubitux: Removed, although I have reservations because of the previously mentioned issue.

libnopegl/src/node_colorstats.c
ubitux: fixed

libnopegl/src/node_renderother.c
ubitux: Sounds overkill but OK. With regards to automation it cannot be done, renderother doesn't work that way: the resources are immediate parameters, not stored in a generic node dict parameters.

libnopegl/src/node_colorstats.c
ubitux: okay, removed

libnopegl/src/node_colorstats.c
mbouron: I'll make a patch to fix the heuristic.

libnopegl/src/node_renderother.c
ubitux: Here is the diff with the `darray` thingy:

```diff
diff --git a/libnopegl/src/node_renderother.c b/libnopegl/src/node_renderother.c
index dcea72135..708045d19 100644
--- a/libnopegl/src/node_renderother.c
+++ b/libnopegl/src/node_renderother.c
@@ -108,7 +108,7 @@ struct render_common {
     int topology;
     const struct geometry *geometry;
     struct darray pipeline_descs;
-    struct ngl_node *extra_draw_node;
+    struct darray draw_resources;
 };

 struct rendercolor_opts {
@@ -594,7 +594,11 @@ static int renderhistogram_init(struct ngl_node *node)
 {
     struct renderhistogram_priv *s = node->priv_data;
     struct renderhistogram_opts *o = node->opts;
-    s->common.extra_draw_node = o->stats;
+
+    ngli_darray_init(&s->common.draw_resources, sizeof(struct ngl_node *), 0);
+    if (!ngli_darray_push(&s->common.draw_resources, &o->stats))
+        return NGL_ERROR_MEMORY;
+
     return init(node, &s->common, &o->common, "source_histogram", source_histogram_frag);
 }

@@ -609,7 +613,11 @@ static int renderwaveform_init(struct ngl_node *node)
 {
     struct renderwaveform_priv *s = node->priv_data;
     struct renderwaveform_opts *o = node->opts;
-    s->common.extra_draw_node = o->stats;
+
+    ngli_darray_init(&s->common.draw_resources, sizeof(struct ngl_node *), 0);
+    if (!ngli_darray_push(&s->common.draw_resources, &o->stats))
+        return NGL_ERROR_MEMORY;
+
     return init(node, &s->common, &o->common, "source_waveform", source_waveform_frag);
 }

@@ -1096,8 +1104,9 @@ static int renderwaveform_prepare(struct ngl_node *node)

 static void renderother_draw(struct ngl_node *node, struct render_common *s, const struct render_common_opts *o)
 {
-    if (s->extra_draw_node)
-        ngli_node_draw(s->extra_draw_node);
+    struct ngl_node **draw_resources = ngli_darray_data(&s->draw_resources);
+    for (int i = 0; i < ngli_darray_count(&s->draw_resources); i++)
+        ngli_node_draw(draw_resources[i]);

     struct ngl_ctx *ctx = node->ctx;
     struct pipeline_desc *descs = ngli_darray_data(&s->pipeline_descs);
@@ -1155,6 +1164,7 @@ static void renderother_uninit(struct ngl_node *node, struct render_common *s)
     ngli_darray_reset(&s->pipeline_descs);
     ngli_buffer_freep(&s->vertices);
     ngli_buffer_freep(&s->uvcoords);
+    ngli_darray_reset(&s->draw_resources);
 }

 #define DECLARE_RENDEROTHER(type, cls_id, cls_name) \
```
Do you confirm you prefer this? I'm not a fan of the init done only in some cases vs the uninit always done, but doing this differently will require unrolling a lot of code.

libnopegl/src/node_colorstats.c
mbouron: Ok, so there is no patch needed as the colorstats node do something a bit hacky regarding the buffer allocatation / initialisation (lazy allocated). In order for the std140 to work here, we also need to set NGLI_BUFFER_USAGE_UNIFORM_BUFFER_BIT flags to the block and buffer usage.

libnopegl/src/node_colorstats.c
```diff
@@ -0,0 +1,413 @@
...
+        int ret = ngli_block_add_field(block, block_layout[i].name, block_layout[i].type, block_layout[i].count);
+        if (ret < 0)
+            return ret;
+    }
+
+    s->blk.buffer = ngli_buffer_create(gpu_ctx);
+    if (!s->blk.buffer)
+        return NGL_ERROR_MEMORY;
+
+    s->blk.usage |= NGLI_BUFFER_USAGE_STORAGE_BUFFER_BIT;
```
mbouron, line 271: Could be `s->blk.usage = NGLI_BUFFER_USAGE_STORAGE_BUFFER_BIT;`

libnopegl/src/node_renderother.c
ubitux: Applied.

libnopegl/src/node_colorstats.c
```diff
@@ -0,0 +1,413 @@
...
+    s->blk.buffer->usage |= NGLI_BUFFER_USAGE_STORAGE_BUFFER_BIT;
```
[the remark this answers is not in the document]
ubitux: Comment added:
```c
    /* Colorstats needs to write into the block so we bind it as SSBO */
    s->blk.usage = NGLI_BUFFER_USAGE_STORAGE_BUFFER_BIT;

    s->blk.buffer = ngli_buffer_create(gpu_ctx);
    if (!s->blk.buffer)
        return NGL_ERROR_MEMORY;

    /*
     * The size of the buffer depends on the texture size, which will only be
     * known in the update callback for textures fed by a Media node. We use a
     * variadic buffer size because don't want to wait for the texture update to
     * compile the shaders and prepare the pipelines.
     *
     * The init and prepare callbacks are still too early for this buffer
     * allocation, but we still need to set its usage for the pipeline bindings.
     */
    s->blk.buffer->usage |= NGLI_BUFFER_USAGE_STORAGE_BUFFER_BIT;
```

libnopegl/src/node_colorstats.c
ubitux: done

mbouron (approved):
LGTM. Nice work :1st_place_medal: . This is a great feature (to say the least).

PR #6, opened 2023-04-08
Displacement node

ubitux:


tests/texture.py
```diff
@@ -80,6 +81,48 @@ def texture_data_unaligned_row(cfg: SceneCfg, h=32):
...
+            ngl_out_pos = ngl_projection_matrix * ngl_modelview_matrix * vec4(ngl_position, 1.0);
+            uv = ngl_uvcoord;
+        }
+        """
+    )
+    frag = dedent(
+        """
+        void main()
+        {
+            float d = (sin(uv.y*10.0+t)+1.0)/2.0*0.1;
```
mbouron, line 107: nit: the expression is a bit too packed.

mbouron:
I wonder if having a dedicated node for this is relevant. How will it scale with other kind of geometry / objects ?

ubitux:
> I wonder if having a dedicated node for this is relevant. How will it scale with other kind of geometry / objects ?

Currently it's effectively only limited to transforming a 2D texture. We need to address how we're going to tackle automatically inserted render target (automatically inside a `Texture2D`, or in `RenderDisplace`, or in other places), how to deal with resolution etc.

mbouron (approved):

PR #7, opened 2023-04-08
Overflow checks

ubitux:
Far from being enough, but that's a start.

### DIFF
--- a/libnopegl/src/hmap.c
+++ b/libnopegl/src/hmap.c
@@ -229,14 +229,22 @@ int ngli_hmap_set(struct hmap *hm, const char *key, void *data)
     if (hm->count * 3 / 4 >= hm->size) {
         struct hmap old_hm = *hm;
 
+#if HAVE_BUILTIN_OVERFLOW
+        size_t new_size;
+        if (__builtin_mul_overflow(hm->size, 2, &new_size))
+            return NGL_ERROR_LIMIT_EXCEEDED;
+#else
+        /* Also includes the realloc overflow check */
         if (hm->size >= 1 << (sizeof(hm->size)*8 - 2))
             return NGL_ERROR_LIMIT_EXCEEDED;
+        size_t new_size = hm->size * 2;
+#endif
 
-        struct bucket *new_buckets = ngli_calloc(hm->size << 1, sizeof(*new_buckets));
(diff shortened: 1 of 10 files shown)


libnopegl/src/hmap.c
```diff
@@ -229,14 +229,22 @@ int ngli_hmap_set(struct hmap *hm, const char *key, void *data)
...
+#if HAVE_BUILTIN_OVERFLOW
+        size_t new_size;
+        if (__builtin_mul_overflow(hm->size, 2, &new_size))
+            return NGL_ERROR_LIMIT_EXCEEDED;
+#else
+        /* Also includes the realloc overflow check */
         if (hm->size >= 1 << (sizeof(hm->size)*8 - 2))
             return NGL_ERROR_LIMIT_EXCEEDED;
+        size_t new_size = hm->size * 2;
+#endif
```
mbouron, line 241: Would it make sense to wrap the overflow check in a dedicated inline function so we don't have to repeat this chunk ? Moreover, if the builtin functions are not supported, the check can be implemented there as a fallback.

ubitux: I didn't find a solution that way because:
1. the builtin does a mutation, so that makes a macro tricky to do
2. if we do a function we need a lot of variants (`int`, `size_t`, etc) but also mixed type (`size_t` with `int`, etc)

mbouron: Would C11 _Generic help here ?

ubitux: `_Generic` is probably mandatory for type inference yes. Should we switch to C11?

It still going to be pretty tricky to write IMO but I can try…

mbouron: This can be addressed later on. I'm just not a huge fan of duplicating this kind of chunks everywhere.

mbouron (approved):

libnopegl/src/hmap.c
ubitux: Yeah me neither, it's ugly. Unfortunately the official checks are going to land in the language in C23.

Note that I started working on the move to `size_t` in most fields so an opportunity might open there at some point.

mbouron: Ok :)

PR #11, opened 2023-04-10
Improved typing annotations in pynopegl

ubitux:


tests/api.py
```diff
@@ -487,3 +488,14 @@ def api_dot(width=320, height=240):
     assert ctx.set_scene(scene) == 0
     assert ctx.dot(0.0) is not None
     assert ctx.dot(1.0) is not None
+
+
+def api_probing():
+    """
+    Simply execute the probing APIs; the result is platform specific so tricky
```
mbouron, line 495: platform/hardware

pynopegl/_pynopegl.pyx
```diff
@@ -465,24 +465,18 @@ def _probe_backends(mode, py_config):
...
-        backend_set.append(dict(
-            id=backend.id,
-            string_id=backend.string_id,
+            caps[cap.string_id] = cap.value
+        backend_set[backend.string_id] = dict(
             name=backend.name,
             is_default=True if backend.is_default else False,
             caps=caps,
-        ))
+        )
```
mbouron, line 467: Maybe this change needs to be mentionned since it changes the API.

libnopegl/src/gen_specs.c
```diff
@@ -19,29 +19,156 @@
...
-            printf("    [\"%s\", \"%s\", \"%s%s%s\"]%s\n", p->key, ngli_params_specs[p->type].name,
-                   (p->flags & NGLI_PARAM_FLAG_ALLOW_LIVE_CHANGE) ? "L" : "",  /* [L]ive */
-                   (p->flags & NGLI_PARAM_FLAG_ALLOW_NODE)        ? "N" : "",  /* [N]ode */
-                   (p->flags & NGLI_PARAM_FLAG_NON_NULL)          ? "M" : "",  /* [M]andatory */
-                   (p + 1)->key ? "," : ""
-            );
+            printf("    {\n");
+            printf("      \"name\": \"%s\",\n", p->key);
+            printf("      \"type\": \"%s\",\n", ngli_params_specs[p->type].name);
+#define D "      \"default\": "
```
mbouron, line 77: Would it make sense to also define a macro for the indentation level ?

libnopegl/src/gen_specs.c
```diff
@@ -19,29 +19,156 @@
...
+            printf("      \"flags\": [");
+            int ihasflag = 0;
+            for (int i = 0; i < NGLI_ARRAY_NB(flag_names); i++) {
+                if (!(p->flags & flag_names[i].flag))
+                    continue;
+                printf("%s\"%s\"", ihasflag ? ", " : "", flag_names[i].name);
+                ihasflag = 1;
+            }
+            printf("],\n");
+            char *esc_desc = escape_json_str(p->desc);
```
mbouron, line 168: D'ont we want to error out in case of alloc failure instead of generating an incomplete spec ?

pynopegl/setup.py
```diff
@@ -106,6 +106,42 @@ def generate(self):
         livectl_info = self._livectl_info()
         return classes + "\n\n" + livectl_info
 
+    @classmethod
+    def _get_combined_type(cls, node_types, wrap_into_union=True):
+        if not node_types:
+            return "Node"
+
+        # We're escaping the node name because we're the node are not declared
```
mbouron, line 114: `because we're the node are not declared`, typo ?

tests/api.py
ubitux: done

pynopegl/_pynopegl.pyx
ubitux: There is an entry added in a commit coming later "pynopegl: pythonize even more the backend probing functions": "Backend probing in `pynopegl` now returns a more Pythonic output" which encapsulate every probing API change

libnopegl/src/gen_specs.c
ubitux: Sure, added

libnopegl/src/gen_specs.c
ubitux: `escape_json_str()` accepts a `NULL` string and makes no distinction with a malloc failure. Since this is a dummy tool I took the liberty to add an `abort()` in case on malloc failure in this function. I hope that's acceptable. I did the same thing in the function itself here:
```c
            case NGLI_PARAM_TYPE_FLAGS: {
                const int v = (int)p->def_value.i32;
                char *s = ngli_params_get_flags_str(p->choices->consts, v);
                if (!s)
                    abort();
```

Alternatively I could add a more advanced error management, but that would imply adding a lot of checks for limited benefit since this is not a program relying on user input.

pynopegl/setup.py
ubitux: Oups yeah, reworded into:
```python
# We're escaping the node name into quotes because the node classes are
# not declared according to their cross dependencies
```

libnopegl/src/gen_specs.c
mbouron: That's fine as is. Thanks.

mbouron (approved):
LGTM, thanks.

PR #12, opened 2023-04-11
build: do not depend on vcpkg for glslang

mbouron:


configure.py
```diff
@@ -125,6 +125,8 @@ def _get_brew_prefix():
 
 
 def _file_chk(path, chksum_hexdigest):
+    if chksum_hexdigest is None:
+        return True
```
ubitux, line 139: I'd much prefer an explicit `"skip"` keyword, otherwise it's likely the next time we add an entry, we end up copying one with no checksum and forget to add it.

Also: "useful" typo in commit description

configure.py
```diff
@@ -75,6 +75,15 @@
         dst_file="SDL2-@VERSION@.zip",
         sha256="446cf6277ff0dd4211e6dc19c1b9015210a72758f53f5034c7e4d6b60e540ecf",
     ),
+    glslang=dict(
+        # Use the legacy master-tot release until the main-tot release provides
+        # Windows binaries
```
ubitux, line 80: `master-tot` and `main-tot` are not exactly "releases"; how about:

> Use the legacy master-tot Windows build until the main-tot one is fixed, or until the glslang project provides Windows builds for their stable releases

mbouron: Changed.

ubitux (approved):
LGTM

PR #14, opened 2023-04-13
api: switch from int to uint32_t for node types

ubitux:
FOURCC are already uint32_t. Various other places in the code play on the aliasing ambiguity between int and uint32_t. This commit make it consistent across the codebase.

As a side effect, -1 is transformed into 0xffffffff. To make it explicit, a new internal define is created to hold this value (NGLI_NODE_NONE).


mbouron (approved):
LGTM.

PR #19, opened 2023-04-23
Scene wrapper

ubitux:


mbouron (approved):
This is nice and it looks good to me !

PR #20, opened 2023-04-26
Workaround a synchronization issue with radeonsi

mbouron:
Fixes https://github.com/nope-project/nope.gl/pull/3 with radeonsi.

### DIFF
--- a/libnopegl/src/backends/gl/glcontext.c
+++ b/libnopegl/src/backends/gl/glcontext.c
@@ -178,12 +178,15 @@ static int glcontext_probe_version(struct glcontext *glcontext)
         glcontext->backend == NGL_BACKEND_OPENGLES ? "ES " : "");
 
     const char *renderer = (const char *)ngli_glGetString(glcontext, GL_RENDERER);
-    LOG(INFO, "OpenGL renderer: %s", renderer ? renderer : "unknown");
+    if (!renderer) {
+        LOG(ERROR, "could not get OpenGL renderer");
+        return NGL_ERROR_BUG;
+    }
+    LOG(INFO, "OpenGL renderer: %s", renderer);
 
-    if (renderer && (
-        strstr(renderer, "llvmpipe") || // Mesa llvmpipe
+    if (strstr(renderer, "llvmpipe") || // Mesa llvmpipe
         strstr(renderer, "softpipe") || // Mesa softpipe
-        strstr(renderer, "SWR"))) {     // Mesa swrast
+        strstr(renderer, "SWR")) {      // Mesa swrast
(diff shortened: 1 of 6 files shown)


libnopegl/src/backends/gl/glcontext.c
```diff
@@ -402,6 +407,43 @@ static int glcontext_probe_settings(struct glcontext *glcontext)
...
+    }
+
+    int mesa_version[3] = {0};
+    const char *mesa = strstr(gl_version, "Mesa");
+    if (mesa) {
+        int ret = sscanf(mesa,
+                         "Mesa %d.%d.%d",
+                         &mesa_version[0],
+                         &mesa_version[1],
+                         &mesa_version[2]);
```
ubitux, line 425: nit: could be just one line, it's not that long, especially if you use `NGLI_ARG_VEC3(&mesa_version)`

libnopegl/src/backends/gl/glcontext.c
```diff
@@ -402,6 +407,43 @@ static int glcontext_probe_settings(struct glcontext *glcontext)
...
+        if (ret != 3) {
+            LOG(ERROR, "could not parse Mesa version (%s)", mesa);
```
ubitux, line 427: nit: `"could not parse Mesa version: \"%s\""`

libnopegl/src/backends/gl/glcontext.c
```diff
@@ -402,6 +407,43 @@ static int glcontext_probe_settings(struct glcontext *glcontext)
...
+            return NGL_ERROR_BUG;
+        }
+        LOG(INFO, "Mesa version: %d.%d.%d", NGLI_ARG_VEC3(mesa_version));
+    }
```
ubitux, line 431: You may want to add in the commit description why you add this unused code, that is I'm assuming because it's going to be used to disable the workaround when the fix is upstreamed. That being said, if the it's actually backported in all maintained releases, we may drop the workaround entirely.

libnopegl/src/backends/gl/glcontext.c
mbouron: Updated (NGLI_ARG_VEC3 cannot be used here though).

libnopegl/src/backends/gl/glcontext.c
mbouron: Done.

libnopegl/src/backends/gl/glcontext.c
mbouron: Done.

ubitux (approved):

PR #21, opened 2023-04-26
colorstats: do not recompute the histogram/summary-scale unless it is necessary

mbouron:


ubitux (commented):
Should be fine; in what situation does this help? Only when the player is in pause and spamming draw events?

mbouron:
> Should be fine; in what situation does this help? Only when the player is in pause and spamming draw events?

It helps when the player is paused but more importantly it  helps in the scope demo scene where the colorstats draw callback is called by the, renderwaveform (parade), renderwaveform and renderhistogram (so instead of computing 3 times the colorstats data, it does it only once).

ubitux:
> > Should be fine; in what situation does this help? Only when the player is in pause and spamming draw events?
> 
> It helps when the player is paused but more importantly it helps in the scope demo scene where the colorstats draw callback is called by the renderwaveform (parade), renderwaveform and renderhistogram (so instead of computing 3 times the colorstats data, it does it only once).

Is this going to work? (note the interlaced compute)

```mermaid
graph

Group -- 0 --> Render0
Group -- 1 --> Compute
Group -- 2 --> Render1
Compute --> Texture
Render0 --> ColorStats
Render1 --> ColorStats
ColorStats --> Texture
```

mbouron:
> > > Should be fine; in what situation does this help? Only when the player is in pause and spamming draw events?
> > 
> > 
> > It helps when the player is paused but more importantly it helps in the scope demo scene where the colorstats draw callback is called by the renderwaveform (parade), renderwaveform and renderhistogram (so instead of computing 3 times the colorstats data, it does it only once).
> 
> Is this going to work? (note the interlaced compute)

Good catch. I'll come up with a better alternative.

mbouron:
Closing for now until a better solution is found.

PR #22, opened 2023-04-26
ci/linux: enable the use of lavapipe to test the Vulkan backend

mbouron:


.github/workflows/ci_linux.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         #
         #         apt-get install python3-venv
         #
-        sudo apt -y install libsdl2-dev python3-venv valgrind spirv-tools glslang-dev libvulkan-dev
+        sudo apt -y install libsdl2-dev python3-venv valgrind spirv-tools glslang-dev libvulkan-dev mesa-vulkan-drivers
```
ubitux, line 40: unrelated but since you're at it: maybe add `spirv-tools glslang-dev libvulkan-dev mesa-vulkan-drivers` to the coverage CI as well? It's disabled though currently…

Also: you may want to add `vk` to `--debug-opts`

ubitux (commented):
Since we don't have a mechanism to force-enable a backend from our configure, can we add something like:

```sh
for b in GL GLES VK; do
    grep -q "BACKEND_$b 1" builddir/libnopegl/config.h
done
```
?

The fail of the `grep` command should abort the CI, you may want to add a `XXX` backend to make sure it does that.

.github/workflows/ci_linux.yml
mbouron: Done.

mbouron:
> Since we don't have a mechanism to force-enable a backend from our configure, can we add something like:
> 
> ```shell
> for b in GL GLES VK; do
>     grep -q "BACKEND_$b 1" builddir/libnopegl/config.h
> done
> ```
> 
> ?
> 
> The fail of the `grep` command should abort the CI, you may want to add a `XXX` backend to make sure it does that.

I would rather add the feature (in a portable way) in our configure in a separe PR, if that's ok with you.

ubitux:
> I would rather add the feature (in a portable way) in our configure in a separe PR, if that's ok with you.

You mean reproducing the enable/disable/auto features mechanism from meson? Maybe we should consider replacing the `configure.py` with a `meson.build` with subprojects and warps at this point. I'm not confident this separate PR reaching `main` in a short time window, but if you want I can make a separate PR with the temporary grep check in the meantime.

mbouron:
> > I would rather add the feature (in a portable way) in our configure in a separe PR, if that's ok with you.
> 
> You mean reproducing the enable/disable/auto features mechanism from meson? Maybe we should consider replacing the `configure.py` with a `meson.build` with subprojects and warps at this point. I'm not confident this separate PR reaching `main` in a short time window, but if you want I can make a separate PR with the temporary grep check in the meantime.

We can just force enable the backends by requesting them explicitely from configure.py. That should be enough and make meson error out if dependencies are not meant.

ubitux (approved):

ubitux:
> EDIT: regarding the temporary check, that will only work for Linux / macOS but not on Windows (and vulkan is enabled there).

Yeah but macos and windows are less important since they're not running the tests; the risk here is to have all the tests from a backend being ignore because ubuntu or khronos may decide to change the way vulkan is being packaged.

mbouron:
> > EDIT: regarding the temporary check, that will only work for Linux / macOS but not on Windows (and vulkan is enabled there).
> 
> Yeah but macos and windows are less important since they're not running the tests; the risk here is to have all the tests from a backend being ignored because ubuntu or khronos may decide to change the way vulkan is being packaged.

I'll submit a patch for the configure.py in the upcoming days (it should not take too much time).

PR #25, opened 2023-05-01
Native C types revamp part 1: -Wconversion (with -Wno-sign-conversion)

ubitux:


mbouron:
Typo in commit msg `tools/desktop: use use single float precision suffix where appropriate`.

PR #26, opened 2023-05-02
graphicconfig: change the scissor type from vec to ivec

mbouron:
The scissor parameter was implemented before the NGLI_PARAM_TYPE_IVEC* types were available. Switching to IVEC4 matches the internal scissor type.

### DIFF
--- a/libnopegl/src/node_graphicconfig.c
+++ b/libnopegl/src/node_graphicconfig.c
@@ -58,16 +58,15 @@ struct graphicconfig_opts {
     int cull_mode;
 
     int scissor_test;
-    float scissor_f[4];
+    int scissor[4];
 };
 
 struct graphicconfig_priv {
     struct graphicstate graphicstate;
     int use_scissor;
-    int scissor[4];
 };
 
-#define DEFAULT_SCISSOR_F {-1.0f, -1.0f, -1.0f, -1.0f}
+#define DEFAULT_SCISSOR {-1, -1, -1, -1}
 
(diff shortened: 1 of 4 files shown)


ubitux (approved):
Nice; can you add "`GraphicConfig.scissor` is now an `ivec4` parameter instead of `vec4`" (or something along these lines in the Changelog?

LGTM otherwise

libnopegl/src/node_graphicconfig.c
```diff
@@ -235,11 +234,8 @@ static int graphicconfig_init(struct ngl_node *node)
         return NGL_ERROR_INVALID_USAGE;
     }
 
-    static const float default_scissor[4] = DEFAULT_SCISSOR_F;
-    s->use_scissor = memcmp(o->scissor_f, default_scissor, sizeof(o->scissor_f));
-    const float *sf = o->scissor_f;
-    const int scissor[4] = {(int)sf[0], (int)sf[1], (int)sf[2], (int)sf[3]};
-    memcpy(s->scissor, scissor, sizeof(s->scissor));
+    static const float default_scissor[4] = DEFAULT_SCISSOR;
```
ubitux, line 237: `static const int32_t`

ubitux (changes requested):

libnopegl/src/node_graphicconfig.c
mbouron: Fixed.

ubitux (approved):

PR #32, opened 2023-05-08
tools/desktop: hardcode mkdir mode value 

mbouron:
Fixes the following warning on macOS:

ngl-desktop.c:585:39: warning: implicit conversion loses integer precision: 'int' to 'mode_t' (aka 'unsigned short') [-Wimplicit-int- conversion]

### DIFF
--- a/ngl-tools/ngl-desktop.c
+++ b/ngl-tools/ngl-desktop.c
@@ -559,7 +559,7 @@ static int setup_network(struct ctx *s)
     return 0;
 }
 
-static int makedirs(const char *path, int mode)
+static int makedirs(const char *path)
 {
     char cur_path[1024];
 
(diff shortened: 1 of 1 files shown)


ubitux (approved):
Please extend the description to add "Changing the makedirs mode parameter from int to mode_t is problematic because of Windows." or something like that.

LGTM otherwise, thx

mbouron:
> Changing the makedirs mode parameter from int to mode_t is problematic because of Windows.

Added 
```
    Changing makedirs() mode parameter from int to mode_t is problematic on
    Windows as it is available on MinGW but not on MSVC. Hardcoding the mkdir
    mode value is the simplest solution and is good enough for now given our
    usage of makedirs().
```
to the commit desc.

PR #37, opened 2023-05-10
Native C types revamp part 3: explicit fixed sized types in various places

ubitux:
Probably far from being enough but a good start.


libnopegl/src/android_imagereader.h
```diff
@@ -57,7 +57,7 @@ void ngli_android_image_freep(struct android_image **sp);
 
 struct android_imagereader;
 
-struct android_imagereader *ngli_android_imagereader_create(struct android_ctx *api, int width, int height, int format, int max_images);
+struct android_imagereader *ngli_android_imagereader_create(struct android_ctx *api, int32_t width, int32_t height, int format, int max_images);
```
mbouron, line 60: I believe the android changes belongs to a separate commit right ?

PR #49, opened 2023-06-04
Limits misc

mbouron:


ubitux (approved):
I'm surprised Vulkan uses `VkDeviceSize` for alignment based fields, but not the sized ones.

LGTM

PR #53, opened 2023-06-05
Pipeline api change

mbouron:


libnopegl/src/pipeline_compat.c
```diff
@@ -228,7 +243,10 @@ void ngli_pipeline_compat_draw_indexed(struct pipeline_compat *s, const struct b
     struct gpu_ctx *gpu_ctx = s->gpu_ctx;
     struct pipeline *pipeline = s->pipeline;
     ngli_gpu_ctx_set_pipeline(gpu_ctx, pipeline);
-    ngli_gpu_ctx_draw_indexed(gpu_ctx, indices, indices_format, nb_indices, nb_instances);
+    for (size_t i = 0; i < s->nb_vertex_buffers; i++)
+        ngli_gpu_ctx_set_vertex_buffer(gpu_ctx, (uint32_t)i, s->vertex_buffers[i]);
+    ngli_gpu_ctx_set_index_buffer(gpu_ctx, indices, indices_format);
```
ubitux, line 248: So we're going toward a stateful approach similar to opengl? Can you elaborate on the overall intent/project?

libnopegl/src/backends/gl/gpu_ctx_gl.c
```diff
@@ -984,6 +984,39 @@ static int gl_get_preferred_depth_stencil_format(struct gpu_ctx *s)
     return NGLI_FORMAT_D24_UNORM_S8_UINT;
 }
 
+static void gl_set_pipeline(struct gpu_ctx *s, struct pipeline *pipeline)
+{
+    struct gpu_ctx_gl *s_priv = (struct gpu_ctx_gl *)s;
+    s_priv->current_pipeline = pipeline;
+}
```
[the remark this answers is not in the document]
mbouron: 1. ngli_gpu_ctx_set_pipeline() currently only store the pipeline, so it would make sense to that, however, I will implement the set_pipeline() mechanism differently in the upcoming PRs and it will be backend specific, for example on Vulkan, set_pipeline() will directly call vkCmdBindPipeline().
2. `pipeline_compat` is an user of the graphics abstraction (not the other way around). Note that there is more change to come to the ngli_gpu_ctx_{draw,draw_indexed,compute}() implementation, ngli_pipeline_*_{draw,draw_indexed,compute} will disappear in the upcoming PRs.

libnopegl/src/pipeline_compat.c
mbouron: My intent is to decorrelate the ressources management from the pipeline to allow binding a pipeline once and execute multiple time with different data in a more saner way than what we do currently (ie: not having the pipeline object encapsulating everything). The API I lean toward is the WGPU API.

PR #58, opened 2023-06-10
Atlas API

ubitux:


libnopegl/src/atlas.c
```diff
@@ -0,0 +1,188 @@
...
+{
+    return 0;
+}
+
+int ngli_atlas_add_bitmap(struct atlas *s, const struct bitmap *bitmap, int32_t *bitmap_id)
+{
+    if (ngli_darray_count(&s->bitmaps) == INT32_MAX)
+        return NGL_ERROR_LIMIT_EXCEEDED;
+
+    struct bitmap copy = {
```
mbouron, line 66: const ?

libnopegl/src/atlas.h
```diff
@@ -0,0 +1,48 @@
...
+#include <stdint.h>
+#include <stdlib.h>
+
+struct ngl_ctx;
+struct atlas;
+
+struct bitmap {
+    uint8_t *buffer;
+    size_t stride;
+    int32_t width, height;
```
mbouron, line 34: uint32_t ?
Not related but I will probably change the texture dimensions to uint32_t.

libnopegl/src/text.c
```diff
@@ -140,7 +137,7 @@ int ngli_text_set_string(struct text *s, const char *str)
         }
 
         const struct char_info chr = {
-            .atlas_id = ngli_drawutils_get_atlas_id(str[i]),
+            .atlas_id = s->ctx->char_map[(uint8_t)str[i]],
```
mbouron, line 140: Why casting to uint8_t here ? Would adding an assert makes more sense ?

libnopegl/src/atlas.h
ubitux: We use `int32_t` everywhere for `width` and `height` so I'd rather keep that consistent

libnopegl/src/atlas.c
ubitux: sure, changed

libnopegl/src/text.c
ubitux: To prevent this warning: `warning: array subscript has type ‘char’ [-Wchar-subscripts]`

libnopegl/src/atlas.h
mbouron: Ok.

libnopegl/src/text.c
mbouron: Also would adding an assert or a check on the value make sense ?

mbouron (approved):
Other than my last remark, LGTM.

libnopegl/src/text.c
ubitux: I'm not sure how it could overflow; `char_map` covers the whole 8-bit range and `str[i]` is 8-bit (just, signed).

mbouron: Ok.

PR #61, opened 2023-06-11
Atlas adjustments

ubitux:


libnopegl/src/atlas.c
```diff
@@ -169,10 +169,17 @@ struct texture *ngli_atlas_get_texture(const struct atlas *s)
     return s->texture;
 }
 
-void ngli_atlas_get_dimensions(const struct atlas *s, int32_t *dim)
```
mbouron, line 172: Typo `altas` in commit desc.

mbouron: `for an appropriate sampling of the bitmap.`  -> `to sample the bitmap correctly.`

libnopegl/src/node_text.c
```diff
@@ -212,8 +211,8 @@ static int update_character_geometries(struct ngl_node *node)
     }
 
     float *transforms = ngli_calloc(text_nbchr, 4 * 4 * sizeof(*transforms));
-    int32_t *atlas_ids = ngli_calloc(text_nbchr, sizeof(*atlas_ids));
-    if (!transforms || !atlas_ids) {
```
mbouron, line 216: You don't need the atlas_id ? even in the future ? (it could be useful for indexing an array from an ubo/ssbo).

ubitux: No plan of using it: all the future parameter will be transmitted individually the same way, through dedicated vertex instance attributes. The benefit over ubo/ssbo is that I don't need to rebuild the whole pipeline when the size changes. If we want to save the pipeline rebuilding with ubo/ssbo, we need to compile a shader with very large arrays for each individual fields. Given how much different fields there will be, it's probably not a good idea to expose a "max size" to the shaders. With instancing, the shader works with individual entries so it's likely more appropriate.

libnopegl/src/atlas.c
ubitux: fixed

mbouron (approved):

PR #62, opened 2023-06-12
Text preparation part 2: drivers and basic layout

ubitux:


libnopegl/src/utils.h
```diff
@@ -115,6 +115,7 @@ char *ngli_strdup(const char *s);
 int64_t ngli_gettime_relative(void);
 char *ngli_asprintf(const char *fmt, ...) ngli_printf_format(1, 2);
 uint32_t ngli_crc32(const char *s);
+uint32_t ngli_crc32_mem(const uint8_t *s, size_t size);
```
mbouron, line 118: Instead of introducing a new function, can we change the prototype of ngli_crc32() instead ?

libnopegl/src/text.c
```diff
@@ -57,17 +57,21 @@ int ngli_text_set_string(struct text *s, const char *str)
     if (ret < 0)
         goto end;
 
+    /* Convert canvas dimension from 26.6 to integer (rounded down) */
+    s->width /= 64;
+    s->height /= 64;
```
mbouron: maybe introduce and use NGLI_I26D6_TO_I32() ?

libnopegl/src/node_text.c
```diff
@@ -325,6 +325,9 @@ static int update_character_geometries(struct ngl_node *node)
             ngli_pipeline_compat_update_attribute(desc->pipeline_compat, 3, s->transforms);
 
             ngli_pipeline_compat_update_attribute(desc->pipeline_compat, 4, s->atlas_coords);
+
+            if (s->text_ctx->cls->mutable_texture)
+                ngli_pipeline_compat_update_texture(desc->pipeline_compat, 0, s->text_ctx->texture);
```
mbouron, line 330: not related but can you add a separate commit to move the ngli_buffer_upload() block below just after the ngli_buffer_init() block above ?

mbouron, line 330: I think you can do that unconditionally, update_character_geometries() is only called on a live change.

libnopegl/src/utils.h
ubitux: Do we really want to pass through the whole string twice? `ngli_crc32(s, strlen(s))` might not be exactly optimal (I'm thinking of the hash map typically)

libnopegl/src/node_text.c
mbouron: Ok, so, isn't there a more appropriate mechanism to signal that the underlying texture has been updated (like a bool in the text ctx, `s->text_ctx->texture_has_changed` or a callback ) ? I believe the atlas won't be recreated every frame, right ?

libnopegl/src/utils.h
mbouron: Ok, makes sense. Should we remame ngli_crc32() to ngli_crc32_str() for consistency ?

libnopegl/src/node_text.c
ubitux: The incoming external font support won't have that much live change events so it's not problematic. Only the builtin one needs to be fast.

Any text change implies an atlas recreation in the case of external font because we are building an atlas per Text instance, which contains only the glyph used in the specified string (otherwise the atlas would be gigantic with certain fonts). If we were to rely on live change update spamming with external fonts, the texture update performance issue addressed here would be the least of our trouble (the re-rasterization of the atlas is going to be much more expensive).

mbouron: Would it make sense to call the field flags and have ..._FLAG_ATLAS_MUTABLE_BIT ? so it is more extensive.

libnopegl/src/text.h
```diff
@@ -56,6 +66,9 @@ struct text {
     int32_t height;
     struct darray chars; // struct char_info
     struct texture *texture;
```
mbouron, line 95: Not related but would it make sense to rename `texture` to `atlas`, `texture` seems too generic here.

libnopegl/src/utils.h
ubitux: Not sure I see the need but if you think that's an improvement I don't mind

libnopegl/src/text.c
ubitux: Added `NGLI_I26D6_TO_I32_TRUNCATED()` and used it (we could have different rounding)

libnopegl/src/text.h
ubitux: Yeah I might but I'm a bit concerned with the potential confusion with the `struct atlas` and its API. I added a commit to rename it to `atlas_texture` instead.

libnopegl/src/node_text.c
ubitux: Sure, changed.

libnopegl/src/text.c
mbouron: Thanks

libnopegl/src/text.h
mbouron: Perfect !

libnopegl/src/node_text.c
mbouron: That's ok as is for now.

mbouron (approved):

PR #64, opened 2023-06-15
Gpu ctx cleanups

mbouron:


libnopegl/src/backends/gl/gpu_ctx_gl.c
```diff
@@ -970,32 +970,27 @@ static int gl_get_preferred_depth_stencil_format(struct gpu_ctx *s)
 
 static void gl_set_pipeline(struct gpu_ctx *s, struct pipeline *pipeline)
 {
-    struct gpu_ctx_gl *s_priv = (struct gpu_ctx_gl *)s;
-    s_priv->current_pipeline = pipeline;
 }
```
ubitux, line 935: Does it make sense to keep this callback?

libnopegl/src/backends/gl/gpu_ctx_gl.c
```diff
@@ -970,32 +970,27 @@ static int gl_get_preferred_depth_stencil_format(struct gpu_ctx *s)
...
 
 static void gl_dispatch(struct gpu_ctx *s, uint32_t nb_group_x, uint32_t nb_group_y, uint32_t nb_group_z)
 {
-    struct gpu_ctx_gl *s_priv = (struct gpu_ctx_gl *)s;
-    struct pipeline *pipeline = s_priv->current_pipeline;
+    struct pipeline *pipeline = s->pipeline;
 
     ngli_assert(pipeline);
     ngli_pipeline_gl_dispatch(pipeline, nb_group_x, nb_group_y, nb_group_z);
```
ubitux, line 958: In both backends these functions are now simple pass-through into the pipeline functions, which means that you could have the `gpu_ctx_class.{draw,draw_indexed,dispatch}` callbacks directly points to their respective `pipeline_{vk,gl}_*` implementations. You would factor out the `pipeline` assert into `gpu_ctx`, and directly call the callbacks on the already stored current pipeline:

```c
void ngli_gpu_ctx_draw(struct gpu_ctx *s, int nb_vertices, int nb_instances)
{
    ngli_assert(s->pipeline);
    s->cls->draw(s->pipeline, nb_vertices, nb_instances);
}

void ngli_gpu_ctx_draw_indexed(struct gpu_ctx *s, int nb_indices, int nb_instances)
{
    ngli_assert(s->pipeline);
    s->cls->draw_indexed(s->pipeline, nb_indices, nb_instances);
}

void ngli_gpu_ctx_dispatch(struct gpu_ctx *s, uint32_t nb_group_x, uint32_t nb_group_y, uint32_t nb_group_z)
{
    ngli_assert(s->pipeline);
    s->cls->dispatch(s->pipeline, nb_group_x, nb_group_y, nb_group_z);
}
```

libnopegl/src/backends/gl/gpu_ctx_gl.c
mbouron: It does, later on, it will directly bind the pipeline (ie: glUseProgram(), and vkCmdBindPipeline()).

libnopegl/src/backends/gl/gpu_ctx_gl.c
mbouron: I prefer to keep it this way for now as I have more changes to the graphics abstraction API. The future change I want to make involves having native calls directly in the backend implementation of draw/draw_indexed/dispatch.

ubitux: Would it make sense to at least move the assert in `gpu_ctx` or that's also counter productive?

mbouron: Oh it makes perfect sense sorry i missed that from your comment (i though it was already the case). Added as a separate commit.

ubitux (approved):

PR #66, opened 2023-06-15
Mvk fix

mbouron:


ubitux (approved):
Is our CI concerned by this fix? You're referring to a build fix for iOS but I don't see any failure, even though I believe we have mvk enabled (maybe only on mac?)

Probably OK anyway

mbouron:
> Is our CI concerned by this fix? You're referring to a build fix for iOS but I don't see any failure, even though I believe we have mvk enabled (maybe only on mac?)

Our macOS CI is concerned by this fix, the macOS build on main fails since the update of MVK.  However, the iOS CI is not concerned as it does support MVK yet.

Regarding the commit related to iOS, it just make the guard consistent for all Apple platforms supporting MVK.

PR #67, opened 2023-06-16
"Basic" external font support

ubitux:


libnopegl/src/text_external.c
```diff
@@ -0,0 +1,749 @@
...
+    hex[(gid) >> 28U & 0xf],    \
+    hex[(gid) >> 24U & 0xf],    \
+    hex[(gid) >> 20U & 0xf],    \
+    hex[(gid) >> 16U & 0xf],    \
+    hex[(gid) >> 12U & 0xf],    \
+    hex[(gid) >>  8U & 0xf],    \
+    hex[(gid) >>  4U & 0xf],    \
+    hex[(gid)        & 0xf],    \
+    '\0'                        \
+}
```
mbouron, line 179: Can we use ngli_crc32_mem() instead ?

libnopegl/src/text_external.c
```diff
@@ -0,0 +1,749 @@
...
+        hb_buffer_set_direction(buffer, HB_DIRECTION_TTB);
+    } else if (text->config.writing_mode == NGLI_TEXT_WRITING_MODE_HORIZONTAL_TB) {
+        /* FriBidi changes the codepoints order from right-to-left where appropriate */
+        hb_buffer_set_direction(buffer, HB_DIRECTION_LTR);
+    } else {
+        ngli_assert(0);
+    }
+
+    // TODO: expose to the user?
+    //hb_buffer_set_script(buffer, script);
```
mbouron, line 337: Can you explain a bit more in the comment how that woul be useful ?
Also can you add a space between `//` and `hb_`

libnopegl/src/text_external.c
```diff
@@ -0,0 +1,749 @@
...
+ * the following bug: https://github.com/fribidi/fribidi/issues/200
+ * This function is pretty much identical without arabic shaping and a few
+ * simplifications due to various unused arguments.
+ */
+static int log2vis(const FriBidiChar *str, int len, FriBidiParType *pbase_dir, FriBidiChar *out_str)
+{
+    int ret = 0;
+
+    // TODO we could re-use a previous allocation or use a stack fallback to
+    // avoid the redundant allocations
```
mbouron, line 460: Can you address this TODO as part of this PR ?

libnopegl/src/text_external.c
```diff
@@ -0,0 +1,749 @@
...
+    struct text_external *s = text->priv_data;
+
+    /* Convert the full string in UTF-8 to Unicode codepoints */
+    const size_t full_len = strlen(str_orig);
+    if (full_len > INT32_MAX)
+        return NGL_ERROR_LIMIT_EXCEEDED;
+    FriBidiChar *codepoints = ngli_calloc(full_len, sizeof(*codepoints));
+    if (!codepoints)
+        return NGL_ERROR_MEMORY;
+    FriBidiStrIndex unicode_len = fribidi_charset_to_unicode(FRIBIDI_CHAR_SET_UTF8, str_orig, (int)full_len, codepoints);
```
mbouron, line 504: `(FriBidiStrIndex)full_len` or `(int32_t)full_len`

libnopegl/src/text_external.c
```diff
@@ -0,0 +1,749 @@
...
+    if (!ngli_darray_push(&s->ft_faces, &ft_face)) {
+        ret = NGL_ERROR_MEMORY;
+        goto end;
+    }
+
+    const int32_t pt_size = 54;
+    const FT_F26Dot6 chr_w = NGLI_I32_TO_I26D6(pt_size);
+    const FT_F26Dot6 chr_h = NGLI_I32_TO_I26D6(pt_size);
+    const FT_UInt h_res = 96;
+    const FT_UInt v_res = 96;
```
mbouron, line 79: Can you add a comment about which size this corresponds to ?

libnopegl/src/text_external.c
```diff
@@ -0,0 +1,749 @@
...
+
+    if (!ngli_darray_push(&s->hb_fonts, &hb_font)) {
+        ret = NGL_ERROR_MEMORY;
+        goto end;
+    }
+
+end:
+    if (ret < 0) {
+        hb_font_destroy(hb_font);
+        FT_Done_Face(ft_face);
```
mbouron, line 101: You can reach this block with a properly initialized/allocated ft_face that has been pushed to s->ft_faces. Destroying it here like that is not enough as s->ft_faces will still reference the pointer.

libnopegl/src/api.c
```diff
@@ -424,6 +424,7 @@ static const char *get_cap_string_id(unsigned cap_id)
     case NGL_CAP_MAX_TEXTURE_DIMENSION_2D:      return "max_texture_dimension_2d";
     case NGL_CAP_MAX_TEXTURE_DIMENSION_3D:      return "max_texture_dimension_3d";
     case NGL_CAP_MAX_TEXTURE_DIMENSION_CUBE:    return "max_texture_dimension_cube";
+    case NGL_CAP_TEXTLIBS:                      return "textlibs";
```
mbouron, line 427: `NGL_CAP_TEXTLIBS` / "textlibs" is, IHMO, not really explicit about what it enables. Would something like `NGL_CAP_TEXT_RENDERING` work ?

libnopegl/src/text_external.c
ubitux: Changed to:
```c
    // TODO: expose to the user? The precise knowledge of the language and/or
    // script may affect the choice of symbols, rules and conventions used
    // during shaping.
    // hb_buffer_set_script(buffer, script);
    // hb_buffer_set_language(buffer, hb_language_from_string("en", -1));
```

libnopegl/src/text_external.c
ubitux: Oops done

libnopegl/src/text_external.c
ubitux: Changed to:
```c
    const FT_F26Dot6 chr_w = NGLI_I32_TO_I26D6(pt_size); // nominal width in 26.6
    const FT_F26Dot6 chr_h = NGLI_I32_TO_I26D6(pt_size); // nominal height in 26.6
    const FT_UInt h_res = 96; // horizontal resolution in dpi
    const FT_UInt v_res = 96; // vertical resolution in dpi
```

libnopegl/src/text_external.c
ubitux: Ah well spotted, error management reworked

libnopegl/src/api.c
ubitux: We do actually have text rendering without external text libraries, so `NGL_CAP_TEXT_RENDERING` wouldn't work.

In general "rendering" doesn't work here, the keyword is libraries, as in "external libraries". `NGL_CAP_TEXT_EXTERNAL_LIBRARIES`?

libnopegl/src/text_external.c
mbouron: Thanks

libnopegl/src/text_external.c
```diff
@@ -0,0 +1,747 @@
...
+    if (end > *pos) {
+        int ret = split_into_runs(text, runs_array, str, len, RUN_TYPE_LINEBREAK, *pos, end);
+        if (ret < 0)
+            return ret;
+        *pos = end;
+    }
+    return 0;
+}
+
+#define STACK_LIST_SIZE 128
```
mbouron, line 440: I guess 128 handle most of the cases ? does that correspond to a line/paragraph of 128 chars ?

libnopegl/src/text_external.c
```diff
@@ -0,0 +1,747 @@
...
+ */
+static int build_text_runs(struct text *text, const char *str_orig, struct darray *runs_array)
+{
+    int ret = 0;
+    struct text_external *s = text->priv_data;
+
+    /* Convert the full string in UTF-8 to Unicode codepoints */
+    const size_t full_len = strlen(str_orig);
+    if (full_len > INT32_MAX)
+        return NGL_ERROR_LIMIT_EXCEEDED;
```
mbouron, line 498: nit: extra line break

libnopegl/src/text_external.c
mbouron: Thanks

libnopegl/src/text_external.c
ubitux: Yup that's exactly that. Incidentally, that's also the value used by fribidi in its `log2vis`

libnopegl/src/text_external.c
ubitux: Done (and moved comment below)

libnopegl/src/text_external.c
mbouron: Ok. Thanks.

libnopegl/src/api.c
mbouron: The idea behind my remark  is that I don't think the end user cares about whether we rely on external libraries or not. He just want to know if the feature that will let him use his fonts is enabled or not. While we indeed support rendering of the builtin font, I don't think it is relevant to do some motions/effects with it.
That being said, `NGL_CAP_TEXT_EXTERNAL_LIBRARIES` is better than `NGL_CAP_TEXTLIBS`.

mbouron (approved):
Other than the last remark regarding the NGL_CAP_* name, LGTM !

libnopegl/src/api.c
ubitux: `NGL_CAP_TEXT_FONTS`?

mbouron: After an offline discussion, we decide to go with `NGL_CAP_TEXT_LIBRARIES`.

PR #70, opened 2023-06-22
Text effects

ubitux:
Depends on https://github.com/nope-project/nope.gl/pull/67


libnopegl/src/text.c
```diff
@@ -127,6 +138,220 @@ int ngli_text_init(struct text *s, const struct text_config *cfg)
...
+{
+    size_t char_id = 0;
+    size_t position = 0;
+
+    const struct char_info_internal *chars = ngli_darray_data(&s->chars_internal);
+    for (size_t i = 0; i < ngli_darray_count(&s->chars_internal); i++) {
+        const struct char_info_internal *c = &chars[i];
+        if (c->tags & NGLI_TEXT_CHAR_TAG_GLYPH)
+            effect->positions[char_id++] = position;
+        position++; // account for all the hidden characters as if they existed
```
mbouron, line 211: as if they exist

libnopegl/src/text.c
```diff
@@ -127,6 +138,220 @@ int ngli_text_init(struct text *s, const struct text_config *cfg)
...
+static void segment_chars_nospace(const struct text *s, struct effect_segmentation *effect)
+{
+    for (size_t i = 0; i < ngli_darray_count(&s->chars); i++)
+        effect->positions[i] = i;
+    effect->total_segments = ngli_darray_count(&s->chars);
+}
+
+static void segment_separator(const struct text *s, struct effect_segmentation *effect, uint32_t mask)
+{
+    int inside_element = 0; // set if we are inside the element
```
mbouron, line 226: nit: the variable is explicit, no need for the associated comment imho

libnopegl/src/text.c
```diff
@@ -127,6 +138,220 @@ int ngli_text_init(struct text *s, const struct text_config *cfg)
...
+        }
+    }
+
+    return 0;
+}
+
+static void destroy_effects_data(struct text *s)
+{
+    /*
+     * effects is not destroyed since its size depends on the number of
```
mbouron, line 340: nit: `text.effects`

libnopegl/src/text.c
```diff
@@ -127,6 +138,220 @@ int ngli_text_init(struct text *s, const struct text_config *cfg)
...
+     * effects (which doesn't change). On the other hand, effects[i].positions
+     * depends on the number of characters, so we're liberating them here
```
mbouron, line 342: `liberating` -> `freeing` ?

libnopegl/src/text.c
```diff
@@ -127,6 +138,220 @@ int ngli_text_init(struct text *s, const struct text_config *cfg)
...
+{
+    for (size_t i = 0; i < s->config.nb_effect_nodes; i++) {
+        struct ngl_node *effect_node = s->config.effect_nodes[i];
+        struct texteffect_opts *effect_opts = effect_node->opts;
+        struct effect_segmentation *effect = &s->effects[i];
+
+        ngli_freep(&effect->positions);
+
+        const size_t nb_chars = ngli_darray_count(&s->chars);
+        effect->positions = ngli_calloc(nb_chars, sizeof(*effect->positions));
```
mbouron, line 300: Can this be reallocated instead of free/malloc ?

libnopegl/src/text.c
```diff
@@ -127,6 +138,220 @@ int ngli_text_init(struct text *s, const struct text_config *cfg)
...
+        case NGLI_TEXT_EFFECT_WORD:         segment_words(s, effect);         break;
+        case NGLI_TEXT_EFFECT_LINE:         segment_lines(s, effect);         break;
+        case NGLI_TEXT_EFFECT_TEXT:         segment_text(s, effect);          break;
+        default:
+            ngli_assert(0);
+        }
+
+        if (effect_opts->random) {
+            /* Build a shuffle map associating a position with another one */
+            size_t *shuffle_map = ngli_calloc(effect->total_segments, sizeof(*shuffle_map));
```
mbouron, line 316: Same here ?

libnopegl/src/text.c
```diff
@@ -209,17 +435,131 @@ int ngli_text_set_string(struct text *s, const char *str)
...
+        .color     = {NGLI_ARG_VEC3(s->config.defaults.color)},
+        .opacity   = s->config.defaults.opacity,
+    };
+
+    /*
+     * Reallocate characters data if the number of characters changed. We could
+     * use a "<" instead of "!=" but we don't want to keep large amount of
+     * memory allocated in case one live change event set a large string which
+     * is later trimed down.
+     */
```
mbouron, line 454: Can  we use a mechanism that allocates every time to the next power of 2 if needed_size > allocated_size ? I'm thinking about a scenario where the user is live editing a node and typing text on the flight, where a realloc would be triggered each type the number of chars increases. Unless i've missed something, the overall memory footprint is rather low compared to what we need when compositing video streams).

libnopegl/src/text.c
```diff
@@ -209,17 +435,131 @@ int ngli_text_set_string(struct text *s, const char *str)
...
+
+            /* Recenter the position in the middle of the character (similar to texture sampling) */
+            const float pos_f = ((float)pos + 0.5f) / (float)nb_elems;
+
+            /* Spacially filter out characters that do not land into the user specified range */
+            if (pos_f < start_pos || pos_f > end_pos)
+                continue;
+
+            /* Interpolate the time of the target, taking into account the overlap */
+            const double t_prv = timescale * (double)pos;
```
mbouron, line 541: prv_t or prev_t

libnopegl/src/text.c
```diff
@@ -209,17 +435,131 @@ int ngli_text_set_string(struct text *s, const char *str)
...
+            const double t_nxt = t_prv + duration;
```
mbouron, line 542: Same, nxt_t, next_t

libnopegl/src/text.h
```diff
@@ -89,14 +98,34 @@ struct text_cls {
...
+struct effect_segmentation {
+    size_t *positions;     // character index (in chars darray) to position in "target unit" (char, word, ...)
+    size_t total_segments; // total number of segment: all values in positions are between [0;total_segments-1]
+};
+
 struct text {
     struct ngl_ctx *ctx;
     struct text_config config;
 
+    /* user exposed */
```
mbouron, line 117: public ?

libnopegl/src/text.h
```diff
@@ -89,14 +98,34 @@ struct text_cls {
...
     int32_t width;
     int32_t height;
     struct darray chars; // struct char_info
     struct texture *atlas_texture;
+    struct text_effects_pointers data_ptrs; // set of effect data pointers (in chars_data)
+
+    /* effects specific */
```
mbouron, line 124: private ?

libnopegl/src/text.c
mbouron: Morever, it happens with the tools ui when the frame number char number increases or decreases, especially when the player seeks back from `duration` to `0`.

libnopegl/src/text.c
ubitux: You sure about that? 🤔

libnopegl/src/text.c
ubitux: Renamed to `inside_target_element` and comment dropped

libnopegl/src/text.c
ubitux: Done

libnopegl/src/text.c
ubitux: Changed

libnopegl/src/text.c
ubitux: Yup, replaced with a realloc

libnopegl/src/text.c
ubitux: Mmh this one is local, not sure it's worth keeping it in the context

libnopegl/src/text.c
ubitux: Ah that's not a bad idea to stitch the allocation to a known power of two. I adjusted the code and extended the comment. (Note: I didn't bother using clz for that)

One thing though: I added a `chars_copy_size`, because we don't need to copy the next-power-of-two amount of data when we have less characters. Making this change made me realize a few short comings in the code, so I made a few other adjustments in that area:

1. `ngli_text_update_effects_defaults()` was not updating the buffers: this was revealed when the re-allocation didn't happen as often as before. It's now filling the default data buffers
2. `chars_data_default` now uses realloc instead of relying on `destroy_effects_data` + `calloc`
3. the default data buffers are now updated only when the number of character changes

We could be smarter to reduce a few more copies but I don't think it's worth the effort for now.

libnopegl/src/text.c
ubitux: renamed to `prev_t`

libnopegl/src/text.c
ubitux: renamed to `next_t`

libnopegl/src/text.h
ubitux: done (I wanted to avoid the term "public" because I associate it with `nopegl.h`)

libnopegl/src/text.h
ubitux: Mmh I prefer specifying that the following block refers to effects specific fields; there are other private fields that do not relate to effects all around

libnopegl/src/text.c
mbouron: The idea is to lower the number of alloc on text updates. It could be left as-is though, we still have the option to rework it later on.

libnopegl/src/text.h
mbouron: Ok.

libnopegl/src/text.c
mbouron: Looks good ! Thanks.

libnopegl/src/text.c
mbouron: Mmm, according to google trad (for what it is worth), it translate `existed`/`exist` to the same tense in french and when going back from french to english, it uses `exist`.

mbouron (approved):
Really nice work :) LGTM !

PR #81, opened 2023-07-04
pipeline_compat: change offset and size type to size_t

mbouron:


libnopegl/src/pipeline.c
```diff
@@ -145,24 +145,18 @@ int ngli_pipeline_update_buffer(struct pipeline *s, int32_t index, const struct
...
         const struct gpu_limits *limits = &s->gpu_ctx->limits;
         if (desc->type == NGLI_TYPE_UNIFORM_BUFFER ||
             desc->type == NGLI_TYPE_UNIFORM_BUFFER_DYNAMIC) {
             ngli_assert(buffer->usage & NGLI_BUFFER_USAGE_UNIFORM_BUFFER_BIT);
-            if (buffer->size > limits->max_uniform_block_size) {
-                LOG(ERROR, "buffer (binding=%d) size (%zu) exceeds max uniform block size (%d)",
-                    desc->binding, buffer->size, limits->max_uniform_block_size);
-                return NGL_ERROR_GRAPHICS_LIMIT_EXCEEDED;
-            }
+            ngli_assert(size <= limits->max_uniform_block_size);
```
ubitux, line 155: Does this mean all the internal users will have to do their own checks? `pass` has some (mixed with automatic type change), but I see other users that have a buffer size depending on user input, will we have to duplicate these checks?

mbouron: Yes. I dropped this commit for now.

ubitux (approved):

ubitux (approved):

PR #84, opened 2023-07-06
Image changes

mbouron:


libnopegl/src/node_texture.c
```diff
@@ -412,6 +412,11 @@ static int handle_media_frame(struct ngl_node *node)
     ngli_image_reset(&s->image);
 
     int ret = ngli_hwmap_map_frame(&s->hwmap, frame, &s->image);
+
+    /* Signal image change on new frame */
+    s->image_rev++;
+    s->image.rev = s->image_rev;
```
ubitux, line 418: Shouldn't it be the other way around?
```c
s->image.rev++; // signal image change on new frame
s->image_rev = s->image.rev; // consider the revision immediately honored
```
?

libnopegl/src/node_colorstats.c
```diff
@@ -416,7 +418,13 @@ static void colorstats_draw(struct ngl_node *node)
     ngli_pipeline_compat_dispatch(s->init.pipeline_compat, s->init.wg_count, 1, 1);
 
     /* Waveform */
-    ngli_pipeline_compat_update_texture_info(s->waveform.pipeline_compat, s->waveform.texture_info);
+    const struct image *image = s->waveform.texture_info->image;
+    ngli_assert(image);
+    if (s->waveform.texture_image_rev != image->rev) {
+        ngli_pipeline_compat_update_texture_info(s->waveform.pipeline_compat, s->waveform.texture_info);
+        s->waveform.texture_image_rev = image->rev;
+    }
```
ubitux, line 426: can't we have the reference revision in the texture info so that `ngli_pipeline_compat_update_texture_info` decides whether to do the update or not?

mbouron: I did that originally but I found that holding this information in the shader crafter is fragile as:
1. this information does not belong to the shader crafter
2. it prevents the crafter from being re-used to create others pipelines.

Yes, having `ngli_pipeline_compat_update_texture_info()` handle that for us would have been great but I believe this needs to be done as part of a big revamp of pipeline_compat (and its goal).

libnopegl/src/node_texture.c
mbouron: Changed to `s->image_rev = s->image.rev++;`

ubitux (commented):
Should be fine; any reason to leave renderother out of this?

mbouron:
> Should be fine; any reason to leave renderother out of this?

I just forgot to update it -> done in a separate commit.

mbouron:
> > Should be fine; any reason to leave renderother out of this?
> 
> I just forgot to update it -> done in a separate commit.

I updated the commits to introduce a texture_map structure (to keep the code consistent with the resource_map structure).

PR #87, opened 2023-07-25
texture: notify texture change on prefetch()/release()

mbouron:
We must notify that the texture has changed after an allocation or de- allocation so the pipelines using the texture take the change into account.


ubitux (approved):
LGTM thanks

PR #88, opened 2023-07-27
Nmd update

mbouron:


ubitux (approved):
Supernice to get rid of the FFmpeg dependency, thanks, LGTM

PR #89, opened 2023-07-27
Misc cleanups + extend path API

ubitux:


libnopegl/src/pipeline_compat.h
```diff
@@ -38,6 +38,7 @@ struct pipeline_compat *ngli_pipeline_compat_create(struct gpu_ctx *gpu_ctx);
 int ngli_pipeline_compat_init(struct pipeline_compat *s, const struct pipeline_compat_params *params);
 int ngli_pipeline_compat_update_vertex_buffer(struct pipeline_compat *s, int32_t index, const struct buffer *buffer);
 int ngli_pipeline_compat_update_uniform(struct pipeline_compat *s, int32_t index, const void *value);
+int ngli_pipeline_compat_update_uniform_count(struct pipeline_compat *s, int32_t index, const void *value, int32_t count);
```
mbouron, line 41: `size_t count` ?

libnopegl/src/block.c
```diff
@@ -197,15 +197,15 @@ int ngli_block_add_field(struct block *s, const char *name, int type, size_t cou
...
+void ngli_block_field_copy_count(const struct block_field *fi, uint8_t * restrict dst, const uint8_t * restrict src, size_t max_count)
 {
     uint8_t *dstp = dst;
     const uint8_t *srcp = src;
 
     if (fi->type == NGLI_TYPE_MAT3) {
         const size_t dst_vec_stride = fi->stride / 3;
         const size_t src_vec_stride = sizes_map[NGLI_TYPE_VEC3];
-        const size_t count = 3 * NGLI_MAX(fi->count, 1);
+        const size_t count = 3 * NGLI_MAX(max_count ? max_count : fi->count, 1);
```
mbouron, line 208: To use the `count` argument name, you can rename the local `count` variable to `n`, so the function prototype is cleared.

libnopegl/src/block.c
```diff
@@ -197,15 +197,15 @@ int ngli_block_add_field(struct block *s, const char *name, int type, size_t cou
     return 0;
 }
 
-void ngli_block_field_copy(const struct block_field *fi, uint8_t * restrict dst, const uint8_t * restrict src)
+void ngli_block_field_copy_count(const struct block_field *fi, uint8_t * restrict dst, const uint8_t * restrict src, size_t max_count)
```
mbouron, line 200: In commit desc: pipeline -> pipeline_compat.

libnopegl/src/block.c
```diff
@@ -228,6 +228,19 @@ void ngli_block_field_copy(const struct block_field *fi, uint8_t *dst, const uin
...
+void ngli_block_fields_copy(const struct block *s, const struct block_field_data *src_array, uint8_t *dst)
+{
+    const struct block_field *fields = ngli_darray_data(&s->fields);
+    for (size_t i = 0; i < ngli_darray_count(&s->fields); i++) {
+        const struct block_field_data *src = &src_array[i];
+        const struct block_field *fi = &fields[i];
+        if (src->count)
+            ngli_block_field_copy_count(fi, dst + fi->offset, src->data, src->count);
+        else
+            ngli_block_field_copy(fi, dst + fi->offset, src->data);
```
mbouron, line 240: I think you can use unconditionnally ngli_block_field_copy_count() here to avoid another branch in ngli_block_field_copy.

libnopegl/src/path.h
```diff
@@ -27,6 +27,18 @@
...
 struct path;
 
+#define SEGMENT_FLAG_NEW_ORIGIN (1 << 0) /* the current segment does not overlap with the previous one */
+#define SEGMENT_FLAG_LINE       (1 << 1) /* the current segment is a simple line (not a curve) */
+
+struct path_segment {
+    float poly_x[4];
+    float poly_y[4];
+    float poly_z[4];
+    int step_start;
```
mbouron, line 42: nit: we might want to switch it to int32_t in a separate commit.

libnopegl/src/path.h
```diff
@@ -50,10 +51,19 @@ int ngli_path_close(struct path *s);
 /* Finalize construction: must be called at the end of the construction */
 int ngli_path_finalize(struct path *s);
 
+/*
+ * Initialization is only required if one wants to evaluate the path at a given
```
mbouron, line 69: Initialize a path. It is only required [...]

libnopegl/src/path.c
```diff
@@ -131,6 +131,11 @@ int ngli_path_bezier3_to(struct path *s, const float *ctl0, const float *ctl1, c
     return add_segment_and_move(s, &segment, to);
 }
 
+int ngli_path_close(struct path *s)
+{
+    return ngli_path_line_to(s, s->origin);
```
mbouron: `s->origin` is introduced in a later commit.

libnopegl/src/path.c
```diff
@@ -417,6 +443,7 @@ void ngli_path_clear(struct path *s)
     ngli_darray_clear(&s->steps);
     ngli_darray_clear(&s->steps_dist);
     memset(s->cursor, 0, sizeof(*s->cursor));
+    memset(s->origin, 0, sizeof(*s->origin));
```
mbouron, line 641: Can you put this above `memset(s->cursor, 0, sizeof(*s->cursor));`

libnopegl/src/path.c
```diff
@@ -142,7 +160,14 @@ int ngli_path_bezier3_to(struct path *s, const float *ctl0, const float *ctl1, c
 
 int ngli_path_close(struct path *s)
 {
-    return ngli_path_line_to(s, s->origin);
+    int ret = ngli_path_line_to(s, s->origin);
```
mbouron, line 178: What happens if:
- the last segment has already the closing flag ? 
- ngli_path_close() is called multiple time ?

libnopegl/src/path.c
```diff
@@ -185,6 +185,103 @@ int ngli_path_close(struct path *s)
...
+    while (*s && strchr(" ,\t\r\n", *s)) /* comma + whitespaces */
+        s++;
+    return s;
+}
+
+static const char *load_coords(float *dst, const char *s, size_t n)
+{
+    for (size_t i = 0; i < n; i++) {
+        s = strip_separators(s);
+        dst[i] = strtof(s, (char **)&s);
```
mbouron, line 199: Are coords limited to float (and not double) ?

libnopegl/src/path.c
```diff
@@ -185,6 +185,103 @@ int ngli_path_close(struct path *s)
...
+            LOG(ERROR, "SVG path command '%c' is currently unsupported", *str);
+            return NGL_ERROR_UNSUPPORTED;
+        } else if (!cmd) {
+            return NGL_ERROR_INVALID_DATA;
+        }
+
+        const char lcmd = cmd | 0x20; // lower case
+
+        if (lcmd == 'z') { // closing current (sub-)path
+            ngli_path_close(s);
```
mbouron, line 239: Missing return value check.

libnopegl/src/path.c
```diff
@@ -282,6 +282,24 @@ int ngli_path_add_svg_path(struct path *s, const char *str)
     return 0;
 }
 
+int ngli_path_add_path(struct path *s, const struct path *add)
```
mbouron, line 285: `const struct path *add` -> `const struct path *path`

libnopegl/src/pipeline_compat.h
ubitux: Oups, old commit, forgot to update this one, fixed.

libnopegl/src/block.c
ubitux: Added an intermediate rename commit and made this change.

libnopegl/src/block.c
ubitux: done

libnopegl/src/block.c
ubitux: Ah good point, changed

libnopegl/src/path.h
ubitux: This one is tricky: it needs to change all the `int` into `i32` in the `path` module, but technically those should be `size_t`. Unfortunately there is a bunch of arithmetic currently, and I'm a bit afraid to break something due to signed operations sometimes.

libnopegl/src/path.h
ubitux: Changed.

libnopegl/src/path.c
ubitux: Ah my bad, move that commit later. Also tested each individual commit to make sure it didn't happen elsewhere.

libnopegl/src/path.c
ubitux: Sure, done

libnopegl/src/path.c
ubitux: I added a commit "path: prevent no-ops when adding lines or bézier curves" which should address these concerns.

libnopegl/src/path.c
ubitux: Yeah doubles are reserved for time (and within animations but only because we want to be able to animate the time)

libnopegl/src/path.c
ubitux: Fixed

libnopegl/src/path.c
ubitux: Renamed

libnopegl/src/path.c
mbouron: I mean on the svg side + is strof() able to read the entire string and truncate appropriately  if needed to perform the conversion from something double to float  ?

mbouron (approved):
Last comment / question from me, otherwhise LGTM.

libnopegl/src/path.c
ubitux: > Unless stated otherwise, numeric values in SVG attributes and in properties that are defined to have an effect on SVG elements must support at least all finite single-precision values supported by the host architecture. 

Source: https://www.w3.org/TR/SVG2/types.html#Precision

See also the first 2 paragraphs here: https://oreillymedia.github.io/Using_SVG/extras/ch08-precision.html

`strtof()` seems to return `HUGE_VALF` in our case if it overflows, whatever that is. I also don't think there is a limit to the length of the string if that's what you had in mind (also, AFAIK `double` don't necessarily have more digits in their strings)

mbouron: Thanks.

PR #98, opened 2023-08-02
ci: wait for Vulkan installers to finish before proceeding further

ubitux:
The installers are GUI based meaning we might get control back before the install ends. This causes the vulkan DLL not to be available during configure, causing all kind of runtime linking issues.

See https://stackoverflow.com/questions/7908000/run-an-application-with-powershell-and-wait-until-it-is-finished for more information on the hack.

Debugged and analyzed with Matthieu Bouron.


mbouron (approved):
LGTM

PR #101, opened 2023-08-03
program_gl_utils: also set block binding for dynamic uniform buffers

mbouron:
Fixes potential binding issues when using dynamic uniform buffers with OpenGL 3.0 contexts.


ubitux (approved):
LGTM thanks

PR #102, opened 2023-08-07
internal: switch to nope.media 11.0.1

mbouron:


libnopegl/meson.build
```diff
@@ -278,7 +278,7 @@ hosts_cfg = {
 
 lib_deps = [
   cc.find_library('m', required: false),
-  dependency('libnopemd', version: '>= 11.0.0'),
+  dependency('libnopemd', version: '>= 11.0.1'),
```
ubitux, line 281: You can drop that we don't have a hard requirement.

ngl-tools/meson.build
```diff
@@ -71,7 +71,7 @@ tool_deps = [
   cc.find_library('m', required: false),
   dependency('libnopegl'),
 ]
-nopemd_dep = dependency('libnopemd', version: '>= 11.0.0')  # for media probing
+nopemd_dep = dependency('libnopemd', version: '>= 11.0.1')  # for media probing
```
ubitux, line 74: Ditto

libnopegl/meson.build
mbouron: Ok.

ngl-tools/meson.build
mbouron: Ok.

ubitux (approved):

PR #104, opened 2023-08-08
text: pack color and opacity in a vec4

ubitux:
This reduces the number of attributes for which the limit can be quite small on some backends/platforms.

### DIFF
--- a/libnopegl/src/text.h
+++ b/libnopegl/src/text.h
@@ -101,8 +101,7 @@ struct text_cls {
 /* Each field points to a contiguous data buffer (1 row per character) */
 struct text_effects_pointers {
     float *transform; // mat4[]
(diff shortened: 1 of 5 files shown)


libnopegl/src/text.h
```diff
@@ -101,8 +101,7 @@ struct text_cls {
 /* Each field points to a contiguous data buffer (1 row per character) */
 struct text_effects_pointers {
     float *transform; // mat4[]
-    float *color;     // vec3[]
-    float *opacity;   // f32[]
+    float *color;     // vec4[] (last field is opacity)
```
mbouron, line 104: field -> component

ubitux: fixed

PR #108, opened 2023-08-09
Mingw/Win cleanups

mbouron:


libnopegl/meson.build
```diff
@@ -104,6 +104,7 @@ if host_system == 'windows'
     add_project_arguments(msvc_args, language: 'c')
   else
     conf_data.set10('TARGET_MINGW_W64', true)
+    add_project_arguments(['-D__USE_MINGW_ANSI_STDIO=1'], language: 'c')
```
ubitux, line 107: don't you need this in `ngl-tools` as well?

libnopegl/src/pthread_compat.h
```diff
@@ -73,7 +73,7 @@ static inline int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
      * conditions.
      * See: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createthread#remarks
      */
-    thread->handle    = (HANDLE)_beginthreadex(NULL, stack_size, pthread_compat_worker, thread, 0, NULL);
+    thread->handle    = (HANDLE)_beginthreadex(NULL, (unsigned)stack_size, pthread_compat_worker, thread, 0, NULL);
```
ubitux, line 76: Note: you probably want this in `nope.media` as well

configure.py
```diff
@@ -727,6 +727,46 @@ def get_env(self):
...
+        f"""\
+        ngli_deactivate() {{
+            deactivate
+            unset NGL_DLL_DIRS
+            unset -f ngli_deactivate
+        }}
+
+        . {activate_path}
+        export NGL_DLL_DIRS="{cfg.bin_path}"
+        """
```
ubitux, line 760: do we really want to set this `NGL_DLL_DIRS` thing on macos and linux?

configure.py
```diff
@@ -728,7 +725,6 @@ def get_env(self):
 
 
 def _build_env_scripts(cfg):
-
```
ubitux: wrong commit

mbouron: Fixed.

configure.py
mbouron: NGL_DLL_DIRS is ignored on platforms other than Windows, so it doesn't hurt to set it unconditionally.

libnopegl/src/pthread_compat.h
mbouron: Indeed and I've updated the commit to add it to ngl-tools.

libnopegl/meson.build
mbouron: Done.

ubitux (approved):

PR #128, opened 2023-09-03
Misc cleanups

ubitux:


README.md
```diff
@@ -38,7 +38,6 @@ and API can change at any time.
 
 ### ⚙️ Discussions and explanations
 
-- [Project history][usr-expl-history]
```
mbouron, line 41: This rationale is not avoid duplication with the nope.website repo right ?

ubitux: Yeah it's annoying to keep both file in sync; I think I actually already adjusted a few things on one side and forgot to backport them already.

The website is public and on git as well so I don't see the benefit of the duplication.

mbouron (approved):

PR #129, opened 2023-09-04
utils/viewer: support export on Windows

mbouron:


pynopegl-utils/pynopegl_utils/viewer/__init__.py
```diff
@@ -251,6 +251,10 @@ def _export_video(self, filename: str, res_index: int, profile_index: int, sampl
         if scene_data is None:
             return
 
+        filename = QUrl(filename).path()  # handle the file:// automatically added by Qt/QML
+        if platform.system() == "Windows" and filename.startswith("/"):
+            filename = filename[1:]
```
ubitux, line 256: Should we add a function for this? We already have 3 instances of it.

And there is also this one which I assume is also broken:
```python
            elif data["type"] == "file" and val is not None:
                val = QUrl(val).path()  # handle the file:// automatically added by Qt/QML
```

pynopegl-utils/pynopegl_utils/viewer/__init__.py
```diff
@@ -291,22 +291,35 @@ def _export_video(self, filename: str, res_index: int, profile_index: int, sampl
...
+            exe = ["ffmpeg"]
+            input = ["pipe:%d" % fd_r]
+
+            if platform.system() == "Windows":
+                import msvcrt
+
+                handle = msvcrt.get_osfhandle(fd_r)
+                os.set_handle_inheritable(handle, True)
+                input = [f"handle:{handle}"]
+                exec = [sys.executable, "-m", "pynopegl_utils.viewer.ffmpeg"]
```
ubitux, line 303: can we name `ffmpeg` to `ffmpeg_win32` or something?

Also, `ffmpeg` or `ffmpeg_bin` sounds better than `exec` to me

pynopegl-utils/pynopegl_utils/viewer/__init__.py
```diff
@@ -291,22 +291,35 @@ def _export_video(self, filename: str, res_index: int, profile_index: int, sampl
...
+            cmd = exec + [
+                "-r", "%d/%d" % fps,
                 "-v", "warning",
                 "-nostats", "-nostdin",
                 "-f", "rawvideo",
                 "-video_size", "%dx%d" % (width, height),
                 "-pixel_format", "rgba",
-                "-i", "pipe:%d" % fd_r,
-            ] + profile.args + [
+                "-i"] + input + profile.args + [
```
ubitux, line 313: I don't think you need an array for the input

pynopegl-utils/pynopegl_utils/viewer/ffmpeg.py
```diff
@@ -0,0 +1,43 @@
...
+#
+
+import platform
+
+if platform.system() != "Windows":
+    raise NotImplementedError
+
+import msvcrt
+import os
+import os.path as op
```
ubitux, line 30: it doesn't look like `op` is used

ubitux (commented):
Would be nice to have a comment somewhere to explain the approach, design, etc

pynopegl-utils/pynopegl_utils/viewer/ffmpeg.py
mbouron: Fixed.

pynopegl-utils/pynopegl_utils/viewer/__init__.py
mbouron: Done.

pynopegl-utils/pynopegl_utils/viewer/__init__.py
mbouron: Done.

pynopegl-utils/pynopegl_utils/viewer/__init__.py
mbouron: Done.

mbouron:
> Would be nice to have a comment somewhere to explain the approach, design, etc

Done in the commit description.

pynopegl-utils/pynopegl_utils/viewer/__init__.py
```diff
@@ -291,22 +291,35 @@ def _export_video(self, filename: str, res_index: int, profile_index: int, sampl
...
+            cmd = ffmpeg + [
+                "-r", "%d/%d" % fps,
                 "-v", "warning",
                 "-nostats", "-nostdin",
                 "-f", "rawvideo",
                 "-video_size", "%dx%d" % (width, height),
                 "-pixel_format", "rgba",
-                "-i", "pipe:%d" % fd_r,
-            ] + profile.args + [
+                "-i", input] + profile.args + [
```
ubitux, line 314: weird indent: can you keep it the way it was before? The `] + profile.args + [` should remain unchanged afaict

pynopegl-utils/pynopegl_utils/viewer/__init__.py
```diff
@@ -39,6 +39,13 @@
 import pynopegl as ngl
 
 
+def _uri_to_path(uri):
+    path = QUrl(uri).path()  # Handle file://
```
ubitux, line 43: Don't want to keep the original comment verbatim? :(

ubitux (commented):
Should be fine aside from these small comments, thanks, that's great :)

pynopegl-utils/pynopegl_utils/viewer/__init__.py
mbouron: Done.

pynopegl-utils/pynopegl_utils/viewer/__init__.py
mbouron: Done. My intent was to drop the comment entirely as it is explicit that we are dealing with uri here.

ubitux (approved):

PR #130, opened 2023-09-04
build: build freetype/harfbuzz/fribidi and enable text support on Windows

mbouron:


configure.py
```diff
@@ -113,6 +128,9 @@ def _get_external_deps(args):
         deps.append("ffmpeg")
         deps.append("sdl2")
         deps.append("glslang")
+        deps.append("harfbuzz")
+        deps.append("freetype")
+        deps.append("fribidi")
```
ubitux, line 133: nit: freetype, harfbuzz, fribidi for consistency with the other order you picked?

ubitux (commented):
I believe It's missing entries in the `clean` rule.

configure.py
mbouron: Done.

mbouron:
> I believe It's missing entries in the `clean` rule.

Done.

ubitux (approved):

PR #131, opened 2023-09-05
Refcount vulkan objects

mbouron:


libnopegl/src/utils.c
```diff
@@ -44,6 +44,22 @@
 #include "utils.h"
 #include "pthread_compat.h"
 
+void ngli_rc_ref(struct ngli_rc *s)
```
ubitux, line 47: `void *ngli_rc_ref` and return `s` so that we can have: `s_priv->attachments_refs[s_priv->nb_attachments] = NGLI_RC_REF(attachment->attachment);` later in this PR?

libnopegl/src/utils.c
```diff
@@ -44,6 +44,22 @@
...
+{
+    s->count++;
+}
+
+void ngli_rc_unrefp(struct ngli_rc **sp) {
```
ubitux, line 52: missing line break

libnopegl/src/utils.h
```diff
@@ -127,6 +127,19 @@
...
 
+typedef void (*ngli_freep_func)(void **sp);
+
+struct ngli_rc {
+    size_t count;
+    ngli_freep_func freep;
+};
+
+#define NGLI_RC_REF(s) ngli_rc_ref((struct ngli_rc *)(s))
+#define NGLI_RC_UNREFP(sp) ngli_rc_unrefp((struct ngli_rc **)(sp))
```
ubitux, line 138: I feel like it makes sense to have `NGLI_RC_NEW(fn) (struct ngli_rc) { .count=1, .freep=fn }` to simplify the next commit

libnopegl/src/bindgroup.h
```diff
@@ -81,16 +83,17 @@ struct buffer_binding {
 };
 
 struct bindgroup_params {
-    const struct bindgroup_layout *layout;
+    struct bindgroup_layout *layout;
```
ubitux, line 86: is this related to this commit? (same below)

libnopegl/src/pipeline.h
```diff
@@ -80,6 +81,7 @@ struct pipeline_params {
 };
 
 struct pipeline {
+    struct ngli_rc rc;
```
ubitux, line 84: Can we have `NGLI_STATIC_ASSERT(rc_is_first, offsetof(struct pipeline, rc) == 0);` for all these structures?

ubitux (changes requested):

libnopegl/src/bindgroup.h
mbouron: Old chunk from a previous iteration. Removed.

libnopegl/src/utils.c
mbouron: Done.

libnopegl/src/utils.c
mbouron: Fixed.

libnopegl/src/utils.h
mbouron: Done. NGLI_RC_CREATE().

libnopegl/src/pipeline.h
mbouron: Done.

libnopegl/src/backends/vk/bindgroup_vk.c
```diff
@@ -289,7 +290,14 @@ int ngli_bindgroup_vk_update_texture(struct bindgroup *s, int32_t index, const s
     struct gpu_ctx_vk *gpu_ctx_vk = (struct gpu_ctx_vk *)s->gpu_ctx;
 
     struct texture_binding_vk *texture_binding = ngli_darray_get(&s_priv->texture_bindings, index);
-    texture_binding->texture = binding->texture ? binding->texture : gpu_ctx_vk->dummy_texture;
+
+    NGLI_RC_UNREFP(&texture_binding->texture);
```
ubitux: Is it possible that we could have the same texture passed down the update_texture? In which case an unref here would destroy the texture and thus the assignment below would be void? Maybe overkill but maybe we should unref it after the new one has been referenced?

libnopegl/src/backends/vk/rendertarget_vk.c
```diff
@@ -328,6 +328,7 @@ static VkResult rendertarget_vk_init(struct rendertarget *s)
             return res;
 
         s_priv->attachments[s_priv->nb_attachments] = view;
+        s_priv->attachments_refs[s_priv->nb_attachments] = NGLI_RC_REF(texture);
```
ubitux, line 331: I'm a bit confused why this is not done above for the color attachment

libnopegl/src/backends/vk/rendertarget_vk.c
```diff
@@ -381,8 +383,10 @@ void ngli_rendertarget_vk_freep(struct rendertarget **sp)
     vkDestroyRenderPass(vk->device, s_priv->render_pass, NULL);
     vkDestroyFramebuffer(vk->device, s_priv->framebuffer, NULL);
 
-    for (uint32_t i = 0; i < s_priv->nb_attachments; i++)
+    for (uint32_t i = 0; i < s_priv->nb_attachments; i++) {
+        NGLI_RC_UNREFP(&s_priv->attachments_refs[i]);
         vkDestroyImageView(vk->device, s_priv->attachments[i], NULL);
```
ubitux, line 388: It looks like of safe but just in case, should the unrefp be after the destroy of the image view (otherwise the image view would be left with dangling reference to the destroyed attachment)

mbouron: Done.

libnopegl/src/backends/vk/bindgroup_vk.c
mbouron: That can't happen as the caller must own a reference to the texture when calling update_texture().
IE:
```
    struct texture *tex = ngli_texture_create();
    ngli_bindgroup_update_texture(b, tex);
    ngli_bindgroup_update_texture(b, tex);
```
The only way to trigger the scenario you described is to somehow call ngli_texture_freep(tex) and still call update_texture() with that dangling pointer.

libnopegl/src/backends/vk/rendertarget_vk.c
mbouron: Me too :) The chunk has simply been forgotten.

PR #133, opened 2023-09-05
Documentation: part 1

ubitux:


libnopegl/scripts/gen-doc.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def _get_node_params_block(nodes, name, params) -> str:
+    s = f"\n## {name}\n\n"
+
+    if not params:
+        return s
+
+    s += "Parameter | Flags | Type | Description | Default\n"
+    s += "--------- | ----- | ---- | ----------- | :-----:\n"
```
mbouron: use `dedent("""...` ?

libnopegl/scripts/gen-doc.py
```diff
@@ -0,0 +1,192 @@
...
+    for param in params:
+        name = param["name"]
+        type = _get_type_str(nodes, param)
+        flags = "".join(f" [`{flag}`](#Parameter-flags)" for flag in param["flags"])
+        desc = param["desc"]
+        default = _get_default_str(param)
+        s += f"`{name}` | {flags} | {type} | {desc} | {default}\n"
+    s += "\n\n"
```
mbouron: Is "\n\n" intended here ? You use a single `\n` in  `_get_parameter_types_block` ?

scripts/gen-doc.py
```diff
@@ -185,7 +185,8 @@ def _main():
     s += _get_parameter_flags_block()
     s += _get_parameter_choices_block(data["choices"])
 
-    print(s.rstrip())
+    with open(sys.argv[2], "w") as f:
```
mbouron, line 188: Do we need to force the encoding here ?

mbouron (approved):

libnopegl/scripts/gen-doc.py
ubitux: done

libnopegl/scripts/gen-doc.py
ubitux: Yeah I wanted to preserve the current output, there are indeed 3 consecutive linebreaks happening in this situation

scripts/gen-doc.py
ubitux: Might be reasonable if we update the doc from Windows. Added.

I also added more forced encoding in the first commit just in case, and in the intermediate commit that still uses stdout (`gen_doc: replace program with a Python script parsing the specs`), I did add the `PYTHONIOENCODING` thing (which later in the history goes away).

libnopegl/scripts/gen-doc.py
mbouron: Ok

PR #138, opened 2023-09-10
Multi-dimensional eval inputs

ubitux:


tests/data.py
```diff
@@ -384,10 +384,13 @@ def data_eval(cfg: SceneCfg):
...
-    b = ngl.UniformFloat(0.3)
-    x = ngl.EvalFloat("sin(a - b + t*4)")
-    x.update_resources(t=t, a=a, b=b)
+    vec = ngl.UniformVec3(value=(0.7, 0.3, 4.0))
+    a = ngl.EvalFloat("vec.0")
+    a.update_resources(vec=vec)
+    b = ngl.EvalFloat("vec.t")
+    a.update_resources(vec=vec)
+    x = ngl.EvalFloat("sin(v.x - v.1 + t*v.p)")
+    x.update_resources(t=t, v=vec)
```
mbouron, line 375: Can you exercise "rgba" as well ?

libnopegl/src/node_eval.c
```diff
@@ -107,6 +127,38 @@ static const struct node_param eval_vec4_params[] = {
 
 NGLI_STATIC_ASSERT(variable_info_is_first, offsetof(struct eval_priv, var) == 0);
 
+static const char * const comp_selectors[] = {"0123", "rgba", "xyzw", "stpq"};
+
+static int register_component_names(struct hmap *vars, const char *base_name, int nb_components, float *ptr)
```
mbouron, line 132: `int nb_components` -> `int32_t` or `size_t`.

libnopegl/src/node_eval.c
```diff
@@ -107,6 +127,38 @@ static const struct node_param eval_vec4_params[] = {
...
+    int ret = 0;
+
+    ngli_assert(nb_components > 1);
+
+    const size_t len = strlen(base_name);
+    if (len >= SIZE_MAX - 3)
+        return NGL_ERROR_LIMIT_EXCEEDED;
+    const size_t size = len + 3;
+    char *name_comp = malloc(size);
+    if (!name_comp)
```
mbouron, line 143: Can't we use an hardcoded max len for the identifier like we do for glsl (and check snprintf return)?

mbouron (approved):
Other than my comments LGTM.

libnopegl/src/node_eval.c
ubitux: Fixed with `size_t`

libnopegl/src/node_eval.c
mbouron: Thanks

PR #143, opened 2023-09-13
GridLayout and misc eval-related changes

ubitux:


doc/usr/howto/eval.md
```diff
@@ -0,0 +1,36 @@
...
+
+```{nope} eval.palette_strip
+:export_type: image
+A palette strip and its grayscale version, interpolated from 2 colors
+```
+
+Here we're using `luma()` to get the greyscale values, and `srgbmix()` to
+interpolate between 2 sRGB colors (interpolation is done in linear space).
+
+Of course, if the interpolation is not required at runtime (that is if it does
```
mbouron, line 32: `that is if it does` -> `ie: the result does`

ubitux: fixed, thanks

mbouron (approved):

PR #145, opened 2023-09-15
Exporter rework

ubitux:


pynopegl-utils/pynopegl_utils/viewer/__init__.py
```diff
@@ -212,6 +212,20 @@ def _select_scene(self, index):
...
+
+        # Check if extension is consistent with format
+        filename = self._export_filename_text.property("text")
+        filename = _uri_to_path(filename)
+        filepath = Path(filename)
+        extension = f".{profile.format}"
+        if filepath.suffix.lower() != extension:
+            filepath = filepath.parent / f"{filepath.stem}{extension}"
+            self._export_filename_text.setProperty("text", filepath.as_posix())
+            self._set_export_filename()
```
mbouron, line 231: Shouldn't we raise an error instead ? Changing the extension automatically, without notifying the user, could lead to the exporter overwriting an existing file.

mbouron (approved):
Except from my comment, LGTM.

pynopegl-utils/pynopegl_utils/viewer/__init__.py
ubitux: Added a message dialog when it happens

mbouron: Thanks. The text box is cropped though, see: https://0x5c.me/screenshots/1695108960_grim.png

PR #149, opened 2023-09-16
From utils to bindings part 1: load_media + move core classes to pynopegl

ubitux:


mbouron (approved):
LGTM. I have a quick question though, the hooks (regarding filepath patching when sending to remote location) are still working ?

ubitux:
> LGTM. I have a quick question though, the hooks (regarding filepath patching when sending to remote location) are still working ?

The full file list was previously patched: `filelist = [m.filename for m in scene_info.medias] + scene_info.files`, it just doesn't include the medias anymore (`filelist = scene_info.files`) but since the `files` list still is there, it's still going through the same logic.

mbouron:
> > LGTM. I have a quick question though, the hooks (regarding filepath patching when sending to remote location) are still working ?
> 
> The full file list was previously patched: `filelist = [m.filename for m in scene_info.medias] + scene_info.files`, it just doesn't include the medias anymore (`filelist = scene_info.files`) but since the `files` list still is there, it's still going through the same logic.

Thanks.

ubitux:
I will push this later (not long before the release) because I don't want to break the examples for people relying on the release to have bad surprises

PR #158, opened 2023-09-26
Texture/RTT resize

mbouron:


libnopegl/src/node_texture.c
```diff
@@ -486,6 +490,67 @@ static int texture_update(struct ngl_node *node, double t)
...
+    s->texture = texture;
+    s->image.params.width = width;
+    s->image.params.height = height;
+    s->image.planes[0] = texture;
+    s->image.rev = s->image_rev++;
+    s->rtt_params = rtt_params;
+    s->rtt_ctx = rtt_ctx;
+
+    return 0;
+fail:
```
ubitux, line 545: nit: can you add a line break before the label?

libnopegl/src/node_texture.c
```diff
@@ -486,6 +490,67 @@ static int texture_update(struct ngl_node *node, double t)
...
+    ngli_texture_freep(&texture);
+    ngli_rtt_freep(&rtt_ctx);
+
+    LOG(ERROR, "failed to resize texture");
```
ubitux, line 549: Can you specify the size?

libnopegl/src/node_rtt.c
```diff
@@ -326,11 +331,123 @@ static int rtt_prefetch(struct ngl_node *node)
...
+        ngli_texture_freep(&texture_priv->texture);
+        texture_priv->texture = depth_texture;
+        texture_priv->image.params.width = width;
+        texture_priv->image.params.height = height;
+        texture_priv->image.planes[0] = depth_texture;
+        texture_priv->image.rev = texture_priv->image_rev++;
+    }
+
+    return 0;
+fail:
```
ubitux, line 430: ditto linebreak

libnopegl/src/node_rtt.c
```diff
@@ -326,11 +331,123 @@ static int rtt_prefetch(struct ngl_node *node)
...
+    for (size_t i = 0; i < o->nb_color_textures; i++)
+        ngli_texture_freep(&textures[i]);
+    ngli_texture_freep(&depth_texture);
+    ngli_rtt_freep(&rtt_ctx);
+
+    LOG(ERROR, "failed to resize rtt");
```
ubitux, line 436: ditto size

libnopegl/src/node_rtt.c
```diff
@@ -326,11 +331,123 @@ static int rtt_prefetch(struct ngl_node *node)
...
+
+    if (o->depth_texture) {
+        const struct rtt_texture_info info = get_rtt_texture_info(o->depth_texture);
+        struct texture_priv *texture_priv = info.texture_priv;
+        ngli_texture_freep(&texture_priv->texture);
+        texture_priv->texture = depth_texture;
+        texture_priv->image.params.width = width;
+        texture_priv->image.params.height = height;
+        texture_priv->image.planes[0] = depth_texture;
+        texture_priv->image.rev = texture_priv->image_rev++;
```
ubitux, line 426: I'm assuming all the other fields (in `image` and `image.params`) are properly initialized and we don't want to reset them?

Should we assert that `image.rev` is `0` at this point? Or should we unref it first?

libnopegl/src/node_texture.c
mbouron: Done.

libnopegl/src/node_rtt.c
mbouron: Done.

libnopegl/src/node_rtt.c
mbouron: Done.

libnopegl/src/node_rtt.c
mbouron: > I'm assuming all the other fields (in `image` and `image.params`) are properly initialized and we don't want to reset them?

Yes.

> 
> Should we assert that `image.rev` is `0` at this point? Or should we unref it first?

`image.rev` is not a ref counter. Its purpose is to provide an unique identifier and a way to notify the texture consumer that the underlying texture has changed.

ubitux (approved):
LGTM thanks

Note: I think the last 2 entries in the changelog (including the one you add here) belong to the "Added" features.

mbouron:
> LGTM thanks
> 
> Note: I think the last 2 entries in the changelog (including the one you add here) belong to the "Added" features.

Moves both entries to the Added section.

PR #170, opened 2023-09-29
QML cleanups & fixes

ubitux:


pynopegl-utils/pynopegl_utils/qml/livectls.py
```diff
@@ -23,8 +23,6 @@
 from dataclasses import dataclass
 from typing import Any, Dict, List
 
-from PySide6.QtGui import QColor
```
mbouron, line 26: nit: in the commit desc:
`and the QML to get and return QColor items. Having that` -> `while the QML rely solely on QColor items` ?

ubitux: Mmh I think it's important to point out that the input *and* output of the QML is Q-typed, while the backend has input *and* output as native. Emphasis is that the model is a conversion bridge.

mbouron (approved):
LGTM ! Thanks.

pynopegl-utils/pynopegl_utils/qml/livectls.py
ubitux: Reworded to:
```
utils/qml: make the model responsible for the QColor conversion

Since the QML is unable to return native types (in the case of the
color, it means a 3-tuple), we want the Model to act as a bridge
between:
- the QML where both input and output are Q-typed
- the backend where both input and output are native Python types (for
  convenience)

Having that conversion happening in the setter and getter (interface
exclusively used by the QML) of the model make it possible for the model
to store the data natively so that the backend only works with native
types.
```

ubitux:
Added changelog entry in the first commit.

pynopegl-utils/pynopegl_utils/qml/livectls.py
mbouron: nit:  ` (in the case of the color, it means a 3-tuple)` -> `(a 3-tuple in this case)`

ubitux: done

PR #171, opened 2023-10-02
Fix & improve backends and capabilities in the Python

ubitux:


mbouron (approved):
Other my comment, LGTM.

pynopegl-utils/pynopegl_utils/tests/cmp.py
```diff
@@ -96,34 +96,30 @@ def __init__(
...
         )
+        ctx = ngl.Context()
+        ret = ctx.configure(ctx_cfg)
         assert ret == 0
 
+        cfg = ngl.SceneCfg(samples=self._samples, clear_color=self._clear_color, backend=ctx_cfg.backend)
+        scene_info = self._scene_func(cfg, **self._scene_kwargs)
+        scene = scene_info.scene
+        duration = scene.duration
+
```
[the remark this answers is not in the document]
ubitux: Done; split in 3 commits

pynopegl-utils/pynopegl_utils/tests/cmp.py
```diff
@@ -96,22 +96,17 @@ def __init__(
...
-
         capture_buffer = bytearray(width * height * 4)
         ctx_cfg = ngl.Config(
             offscreen=True,
             width=width,
             height=height,
-            backend=cfg.backend if backend else ngl.Backend.AUTO,
-            samples=cfg.samples,
-            clear_color=cfg.clear_color,
+            backend=get_backend(backend) if backend else ngl.Backend.AUTO,
```
mbouron, line 107: I think this does not belong to the commit.

pynopegl-utils/pynopegl_utils/config.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, module_pkgname):
             "framerate": (60, 1),
             "log_level": "info",
             "clear_color": (0.0, 0.0, 0.0, 1.0),
-            "backend": "opengl",
+            "backend": next(k.name.lower() for k, v in ngl.get_backends().items() if v["is_default"]),
```
mbouron, line 80: Have you tested to change the default backend internally (and pick vulkan for example) , does it work as intented ?

pynopegl-utils/pynopegl_utils/tests/cmp.py
ubitux: I don't have `cfg` available anymore, so if I want to create an intermediate commit, I need to do something like that:
```python
         backend = os.environ.get("BACKEND")
         if backend:
-            cfg.backend = get_backend(backend)
+            backend = get_backend(backend)
```
But that introduces a mismatch of type (could be `str`, `None` or `Backend`), which would then be rechecked again with `backend=backend if backend else ngl.Backend.AUTO` which is not a good intermediate state I feel because it's pretty confusing what `backend` can be.

pynopegl-utils/pynopegl_utils/config.py
ubitux: Changed the default, removed the config:

2023-10-11-132142-aeraY0il

mbouron: Thanks.

PR #175, opened 2023-10-11
doc: provide both main and latest stable versions of the documentation

ubitux:
Note that the redirecting message will only be visible starting next release.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -27,6 +27,8 @@ jobs:
     steps:
     - name: Checkout sources
       uses: actions/checkout@v3
+      with:
+        fetch-tags: true
 
     - name: Install dependencies
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/doc.yml
```diff
@@ -44,15 +46,32 @@ jobs:
         sudo apt -y install libfreetype-dev libharfbuzz-dev libfribidi-dev
         sudo apt -y install graphviz
 
-    - name: Build
+    - name: Build (main)
+      run: |
+        ./configure.py --buildtype debug --debug-opts gl vk gpu_capture
```
mbouron, line 51: Would it possible to factorize the build args into a env variable to avoid duplication ?

mbouron (approved):
Other than my comment, LGTM.

.github/workflows/doc.yml
mbouron: After an offline discussion, we will do that later as it requires a new round of tests.

PR #177, opened 2023-10-23
Hwconv fixes

mbouron:
The crash can be reproduced with ngl-player -d 0 <media> when using a hwaccel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Versioning](https://semver.org/spec/v2.0.0.html) for `libnopegl`.
 
 ### Fixed
 - Moving the split position in `ngl-diff`
+- Crash in the hwconv module when direct rendering is not possible/enabled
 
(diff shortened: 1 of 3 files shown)


mbouron:
> A bit weird to have to mutate the data held by pgcraft. Maybe the image should be de-correlated from the texture info.
> 
> I'm seeing this comment in pgcraft: `FIXME: this means the internally exposed pgcraft_texture_info contains many unnecessary stuff for our users of the pgcraft API.` so I feel like we're missing an intermediate structure/layer or something here.

The comment in pgcraft is more about a shortcut that we took a while ago to make the code "simpler" by not introducing another structure and in turn exposing too much information.

Regarding the current api of the texture abstraction, I agree it's weird. Also note that we are already using this pattern in other places where where we mutate the image held inside the pgcraft_texture_info.

Anyway, I'll propose a patchset to address both issues.

PR #192, opened 2023-11-10
Pgcraft pipeline misc

mbouron:


configure.py
```diff
@@ -843,6 +846,7 @@ def _run():
     parser.add_argument("-p", "--venv-path", default=op.join(_ROOTDIR, "venv"), help="Virtual environment directory")
     parser.add_argument("--buildtype", choices=("release", "debug"), default="release", help="Build type")
     parser.add_argument("--coverage", action="store_true", help="Code coverage")
+    parser.add_argument("--sanitize-address", action="store_true", help="Enable address sanitizer")
```
ubitux, line 849: can you use the `--sanitize=xxx` form for the argument instead so we can add the `undefined` as well at some point (and eventually threading)

libnopegl/src/pgcraft.c
```diff
@@ -1428,11 +1427,11 @@ struct pgcraft *ngli_pgcraft_create(struct ngl_ctx *ctx)
...
 
     ngli_darray_init(&s->pipeline_info.data.textures,   sizeof(struct texture *), 0);
-    ngli_darray_init(&s->pipeline_info.data.buffers,    sizeof(struct buffer *),  0);
+    ngli_darray_init(&s->pipeline_info.data.buffers,    sizeof(struct buffer_binding),  0);
     ngli_darray_init(&s->pipeline_info.data.vertex_buffers, sizeof(struct buffer *),  0);
 
     ngli_darray_init(&s->filtered_pipeline_info.data.textures,   sizeof(struct texture *), 0);
-    ngli_darray_init(&s->filtered_pipeline_info.data.buffers,    sizeof(struct buffer *),  0);
+    ngli_darray_init(&s->filtered_pipeline_info.data.buffers,    sizeof(struct buffer_binding),  0);
     ngli_darray_init(&s->filtered_pipeline_info.data.vertex_buffers, sizeof(struct buffer *),  0);
```
ubitux, line 1435: it probably makes sense to rename `buffers`, because we now have `buffers` and `vertex_buffers` with different types

libnopegl/src/pipeline_compat.c
```diff
@@ -401,6 +401,58 @@ int ngli_pipeline_compat_update_buffer(struct pipeline_compat *s, int32_t index,
...
+        return 0;
+
+    /* Otherwise, increase our bindgroup pool */
+    size_t nb = s->nb_bindgroups;
+    size_t nb_bindgroups = s->nb_bindgroups * 2;
+    struct bindgroup **bindgroups = ngli_realloc(s->bindgroups, nb_bindgroups, sizeof(*bindgroups));
+    if (!bindgroups)
+        return NGL_ERROR_MEMORY;
+
+    memset(bindgroups + s->nb_bindgroups, 0, s->nb_bindgroups * sizeof(*bindgroups));
```
ubitux, line 422: `darray`?

libnopegl/src/pipeline_compat.c
```diff
@@ -401,6 +401,58 @@ int ngli_pipeline_compat_update_buffer(struct pipeline_compat *s, int32_t index,
...
+        };
+
+        int ret = ngli_bindgroup_init(s->bindgroups[i], &params);
+        if (ret < 0)
+            return ret;
+    }
+
+    /* And select the first bindgroup from our newly allocated bindgroups */
+    s->cur_bindgroup = s->bindgroups[s->next_bindgroup++];
+    s->next_bindgroup %= s->nb_bindgroups;
```
ubitux, line 448: A bit weird to have this done twice in this codepath (first occurrence at the beginning)

libnopegl/src/node_colorstats.c
```diff
@@ -255,24 +255,17 @@ static int init_block(struct colorstats_priv *s, struct gpu_ctx *gpu_ctx)
     struct block *block = &s->blk.block;
     ngli_block_init(gpu_ctx, block, NGLI_BLOCK_LAYOUT_STD430);
 
-    /* Set block fields */
-    static const struct {
-        const char *name;
-        int type;
-        size_t count;
-    } block_layout[] = {
+    const struct block_field block_fields[] = {
```
ubitux, line 257: I believe it can remain `static`

mbouron: Done.

configure.py
mbouron: Done. The choices are taken directly from the meson documentation:
```
The value of b_sanitize can be one of: none, address, thread, undefined, memory, leak, address,undefined, but note that some compilers might not support all of them. For example Visual Studio only supports the address sanitizer.
```

libnopegl/src/pipeline_compat.c
mbouron: I'm currently not using any darray around here so I would rather keep it as is.

libnopegl/src/pipeline_compat.c
mbouron: I tried to split the logic in another function but it does not improve the overall readability and adds an indirection.

configure.py
ubitux: lol weird

libnopegl/src/pipeline_compat.c
mbouron: Changed to darray and simplified the logic.

libnopegl/src/pipeline_compat.c
mbouron: Changed to darray and simplified the logic.

libnopegl/src/pgcraft.c
mbouron: I would prefer to keep its name short. Also it aligns with the pipeline_compat.resources field name.

libnopegl/src/pipeline_compat.c
```diff
@@ -122,6 +121,51 @@ static int init_blocks_buffers(struct pipeline_compat *s, const struct pipeline_
     return 0;
 }
 
+static void free_bindgroup(void *user_arg, void *data)
+{
+    struct bindgroup *bindgroup = data;
+    ngli_bindgroup_freep(&bindgroup);
+}
+
```
ubitux, line 129: I think this is incorrect; I believe you need this:
```diff
diff --git a/libnopegl/src/pipeline_compat.c b/libnopegl/src/pipeline_compat.c
index 43f261609..80692787f 100644
--- a/libnopegl/src/pipeline_compat.c
+++ b/libnopegl/src/pipeline_compat.c
@@ -123,8 +123,8 @@ static int init_blocks_buffers(struct pipeline_compat *s, const struct pipeline_

 static void free_bindgroup(void *user_arg, void *data)
 {
-    struct bindgroup *bindgroup = data;
-    ngli_bindgroup_freep(&bindgroup);
+    struct bindgroup **bindgroupp = data;
+    ngli_bindgroup_freep(bindgroupp);
 }

 static int grow_bindgroup_array(struct pipeline_compat *s)
```

mbouron: Thanks for pointing that out. Fixed.

ubitux (approved):

PR #193, opened 2023-11-11
Darray free

ubitux:


libnopegl/src/hmap.c
```diff
@@ -38,13 +38,13 @@ struct hmap {
     size_t size;
     size_t mask;
     size_t count; // total number of entries
-    user_free_func_type user_free_func;
+    hmap_user_free_func_type user_free_func;
```
mbouron, line 41: Can we define `ngli_user_free_func_type` (or even ngli_user_freep_func_type if that makes sense , that would in line with our current `ngli_freep_func_type`) in `utils.h` and use it for both hmap and darray ?

libnopegl/src/darray.c
```diff
@@ -128,13 +137,25 @@ void *ngli_darray_get(const struct darray *darray, size_t index)
     return darray->data + index * darray->element_size;
 }
 
+static void invalidate_range(struct darray *darray, size_t index, size_t count)
+{
+    if (!darray->user_free_func)
+        return;
+    const size_t end = index + count;
```
mbouron, line 144: You ll probably want the same check as it `remove_range`: `ngli_assert(end <= darray->count);`

libnopegl/src/backends/vk/bindgroup_vk.c
```diff
@@ -241,6 +258,9 @@ int ngli_bindgroup_vk_init(struct bindgroup *s, const struct bindgroup_params *p
     ngli_darray_init(&s_priv->texture_bindings, sizeof(struct texture_binding_vk), 0);
     ngli_darray_init(&s_priv->buffer_bindings, sizeof(struct buffer_binding_vk), 0);
 
+    ngli_darray_set_free_func(&s_priv->texture_bindings, unref_texture_binding, NULL);
+    ngli_darray_set_free_func(&s_priv->buffer_bindings, unref_buffer_binding, NULL);
```
mbouron, line 262: nit: personal preference, i would pack the init + set_free_func per array.

libnopegl/src/hud.c
```diff
@@ -952,9 +952,21 @@ static int create_widget(struct hud *s, enum widget_type type, const void *user_
     return 0;
 }
 
+static void delete_widget(void *user_arg, void *data)
```
mbouron, line 955: `free_widget` ?

mbouron (approved):

libnopegl/src/hud.c
ubitux: it's doing both an uninit and then a free so I wanted a common term 🤷

mbouron: `reset` then ? we don't `delete` in the rest of the code base

libnopegl/src/hmap.c
ubitux: done

libnopegl/src/darray.c
ubitux: added

libnopegl/src/hud.c
ubitux: do you want me to rename all the other darray callbacks as well?

mbouron: Nope. AFAIK, delete_ is only used here.

ubitux: done

PR #194, opened 2023-11-11
Misc

mbouron:


libnopegl/meson.build
```diff
@@ -130,6 +130,7 @@ project_args = cc.get_supported_arguments([
   '-Wno-sign-conversion',
   '-fmacro-prefix-map=@0@/='.format(trim_prefix),
   '-fno-math-errno',
+  '-Wundef',
```
ubitux, line 133: probably makes sense in ngl tools as well?

Also, can you add it below the last `-W*` so that warning flags are together?

mbouron: Done

ubitux (approved):

PR #198, opened 2023-11-16
tools/desktop: fix backend probing

ubitux:
Regression since e30dbd400eed593e29c27cd73a426a3f27ffc024.


mbouron (approved):
Sorry for the regression. Thanks.

PR #200, opened 2023-11-17
Misc fixes

mbouron:


libnopegl/src/node_media.c
```diff
@@ -331,7 +331,6 @@ static int media_update(struct ngl_node *node, double t)
                     pix_fmt_str ? pix_fmt_str : "unknown", frame->pix_fmt);
                 return NGL_ERROR_BUG;
             }
-            pix_fmt_str = "audio";
```
ubitux, line 328: it's needed for the trace log below

libnopegl/src/text.c
```diff
@@ -318,6 +318,9 @@ static uint64_t prng64_next(uint64_t *state)
 
 static void shuffle(uint64_t *rng_state, size_t *positions, size_t n)
 {
+    if (n == 0)
+        return;
```
ubitux, line 322: 0 segment is likely going to cause divisions by zero and all sort of problems. It might be possible to trigger it so I'll investigate, but please drop this commit since it should be an assert that deep in the codepath

mbouron: Up to you, dropped.

libnopegl/src/node_media.c
mbouron: Dropped.

ubitux (approved):

PR #202, opened 2023-11-20
Scene setup nodes

ubitux:
Depends on https://github.com/NopeForge/nope.gl/pull/191


libnopegl/src/nodes.c
```diff
@@ -48,6 +48,7 @@ NGLI_STATIC_ASSERT(node_anim_vec_flt,    NGL_NODE_ANIMATEDVEC4     - NGL_NODE_AN
 
 extern const struct param_specs ngli_params_specs[];
 
+/* Warning: there is an assumption that these parameters do not include any nodes */
```
mbouron, line 51: nit: something `The common node parameters *must* not include any node-based parameter`

mbouron: Maybe we can put an assert when processing the root node ? (so it will be run once) ?

libnopegl/src/internal.h
```diff
@@ -702,6 +702,9 @@ void *ngli_node_get_data_ptr(struct ngl_node *var_node, void *data_fallback);
 int ngli_prepare_draw(struct ngl_ctx *s, double t);
 void ngli_node_draw(struct ngl_node *node);
 
+typedef int (*ngli_children_func_type)(void *user_arg, struct ngl_node *node, struct ngl_node *child);
+int ngli_node_children_apply_func(ngli_children_func_type func, void *user_arg, struct ngl_node *node);
```
mbouron: Can we have the argument in the following order (and node renamed to parent for the func type):
typedef int (*ngli_children_func_type)(struct ngl_node *parent, struct ngl_node *child, void *user_arg);`
`int ngli_node_children_apply_func(struct ngl_node *node, ngli_children_func_type func, void *user_arg);`
?

mbouron: Also maybe renamed the function to `ngli_node_children_func_type` ?

libnopegl/src/api.c
```diff
@@ -249,6 +249,12 @@ int ngli_ctx_set_scene(struct ngl_ctx *s, struct ngl_scene *scene)
             goto fail;
         }
 
+        if (scene->params.root->ctx) {
+            LOG(ERROR, "the specified scene is already associated with a context");
```
mbouron, line 253: +rendering ?

libnopegl/src/nodes.c
```diff
@@ -291,53 +291,51 @@ static int node_init(struct ngl_node *node)
     return 0;
 }
 
-static int node_set_ctx(struct ngl_node *node, struct ngl_ctx *ctx, struct ngl_ctx *pctx)
+static int node_set_ctx(struct ngl_node *node, struct ngl_ctx *ctx)
```
mbouron, line 243: Regarding the commit msg:  `nodes: split node_set_ctx() into two separate function`

libnopegl/src/nodes.c
```diff
@@ -291,21 +291,6 @@ static int node_init(struct ngl_node *node)
     return 0;
 }
 
-static int node_set_ctx(struct ngl_node *node, struct ngl_ctx *ctx, struct ngl_ctx *pctx);
```
mbouron: nit: ngli_node_children_apply_func() in commit msg

mbouron (approved):

tests/api.py
```diff
@@ -151,27 +151,30 @@ def api_ctx_ownership():
...
+    shared_geometry = ngl.Quad()
+    scene0 = ngl.Scene.from_params(ngl.RenderColor(geometry=shared_geometry))
+    try:
+        scene1 = ngl.Scene.from_params(ngl.RenderColor(geometry=shared_geometry))
+    except Exception:
+        pass
+    else:
+        del scene1
+        assert False
+    del scene0
```
mbouron, line 177: While at it, can you also test with a ngl context attached (in various scenarios) ?
- sharing the graph
- sharing a subpart of the graph
- before and after detaching the initial ngl ctx / scene

libnopegl/src/nodes.c
ubitux: - Comment updated.
- Added a dedicated check function in `gen_specs.c` (see also the added associated commit `gen_specs: remove ret variable shadowing`)

libnopegl/src/internal.h
ubitux: - Fixed the node/child vs parent/node confusion
- `user_arg` is usually first in the codebase (rationale being that it's a context)
- I don't understand the `ngli_node_children_func_type` function rename; `_func_type` refers to a typedef

libnopegl/src/api.c
ubitux: Added

libnopegl/src/nodes.c
ubitux: Added `()` and the `s`

libnopegl/src/nodes.c
ubitux: Done in 2-3 other commits

tests/api.py
ubitux: Sharing a subpart of the graph is already tested at the scene level, so it's not possible at the context level, and it also implies sharing the whole graph.

I'm not sure what you mean by before and after detaching but I added a test to transfer a scene from one context to another.

libnopegl/src/internal.h
mbouron: Thanks. Regarding `user_arg` I got confused by the fact that we placed `user_arg` in the recently introduced callbacks setters (see `ngli_darray_set_free_func`, `ngli_darray_set_free_func`).

Regarding the rename it's about: `ngli_children_func_type` -> `ngli_node_children_func_type`.

tests/api.py
mbouron: To clarify a bit: if a scene/graph is associated with a rendering context (and is up and running), we want to make sure (and I mean that we need a test for this scenario) that any intervention from the user with the graph or the subgraph won't mess up the rendering context.

libnopegl/src/internal.h
ubitux: The function is renamed to `children_apply_func` so I guess I can ignore that? :D

mbouron: Ok.

tests/api.py
ubitux: Added a dedicated commit: `nodes: prevent graph structure changes after scene association`

mbouron: Ok but, what i meant is more about, you associate a graph with a rendering context, you test that draw succeed. Then, you try to ref the first graph from another graph, try to associate it with a scene, then a rendering context if the previous operation succeeded (i believe it should not so skip this step) then you check again that the first rendering context is still working by checking that a draw succeeded. The goal is to ensure that the error path in the attach/init functions does not mess with the first graph.

ubitux: Added `api_scene_resilience` test, but note that I couldn't work with the original scene object directly due to limitation in the Python binding (`.from_params()` is a class constructor and can't be re-used on an existing instance)

libnopegl/src/nodes.c
```diff
@@ -747,6 +747,10 @@ static int param_add(struct ngl_node *node, const char *key, size_t nb_elems, vo
 int ngl_node_param_add_nodes(struct ngl_node *node, const char *key,
                              size_t nb_nodes, struct ngl_node **nodes)
 {
+    if (node->scene) {
+        LOG(ERROR, "the nodes graph cannot be extended after being associated with a scene");
+        return NGL_ERROR_INVALID_USAGE;
+    }
```
mbouron, line 545: Is this a new limitation introduced by this commit or by the commit that introduced the scene association ?

ubitux: It's not allowed started from `scene: check for node sharing within the scene context`, but:
1. I can't move that commit before (since the `scene` pointer is not available yet)
2. Squashing it into that commit would make the commit quite indigestible so I prefer to keep it separate; the implication being that a few sanity checks are missing for a few commits, which I think is acceptable?

I added this to the description of that commit:
```
These operations are undefined behaviour since the introduction of the
scene pointer in the commit "check for node sharing within the scene
context".
```

tests/api.py
mbouron: Ok. Thanks.

PR #206, opened 2023-11-21
Hmap u64

ubitux:


libnopegl/src/hmap.c
```diff
@@ -54,7 +64,27 @@ void ngli_hmap_set_free_func(struct hmap *hm, ngli_user_free_func_type user_free
 #define NO_REF (struct hmap_ref){.bucket_id = -1}
 #define HAS_REF(ref) ((ref).bucket_id != -1)
 
-struct hmap *ngli_hmap_create(void)
+static uint32_t key_hash_str(union hmap_key x) { return ngli_crc32(x.str); }
+static uint32_t key_hash_u64(union hmap_key x) { return ngli_crc32_mem(x.u8_8, sizeof(x.u8_8)); }
```
mbouron, line 68: Personal preference but ... can  you avoid interleaving the declaration ? it's hard to follow, especially if there is allocation involved.

mbouron (approved):
Except from my comment, LGTM.

libnopegl/src/hmap.c
```diff
@@ -118,17 +118,17 @@ static struct hmap_ref ref_from_entry(const struct hmap *hm, const struct hmap_e
...
 }
 
 static struct hmap_entry *fixed_entry_from_ref(const struct hmap *hm, struct hmap_ref ref,
-                                               struct hmap_ref removed, int current_id)
+                                               struct hmap_ref removed, size_t current_id)
 {
     if (!HAS_REF(ref))
         return NULL;
-    const int need_fix =
+    const size_t need_fix =
```
mbouron, line 131: It looks like need_fix is a boolean. Isn't better to cast it later on ? or simplify use a `if` statement and sub explictely by `1`.

libnopegl/src/hmap.c
ubitux: Batching functions with the same role is IMO way more readable and less error prone than the following:

```c
static uint32_t key_hash_str(union hmap_key x) { return ngli_crc32(x.str); }
static int key_cmp_str(union hmap_key a, union hmap_key b) { return strcmp(a.str, b.str); }
static union hmap_key key_dup_str(union hmap_key x) { return (union hmap_key){.str=ngli_strdup(x.str)}; }
static int key_check_str(union hmap_key x) { return !!x.str; }
static void key_free_str(union hmap_key x) { ngli_free(x.str); }

static uint32_t key_hash_u64(union hmap_key x) { return ngli_crc32_mem(x.u8_8, sizeof(x.u8_8)); }
static int key_cmp_u64(union hmap_key a, union hmap_key b) { return a.u64 != b.u64; }
static union hmap_key key_dup_u64(union hmap_key x) { return x; }
static int key_check_u64(union hmap_key x) { return 1; }
static void key_free_u64(union hmap_key x) { }
```

libnopegl/src/hmap.c
ubitux: Yep, done with a cast

libnopegl/src/hmap.c
mbouron: Again it's personal preference. Since you are the main contributor to this module, you can keep it as-is.

mbouron (approved):

PR #207, opened 2023-11-21
utils: use __typeof__() in NGLI_SWAP()

mbouron:


mbouron:
Closing for now, MSVC does not support __typeof__, so better wait for c23 !

PR #215, opened 2024-01-16
Selector filter

ubitux:


tests/filter.py
```diff
@@ -132,6 +134,32 @@ def filter_saturation(_):
...
+    )
+
+
+@test_cuepoints(points=_CUEPOINTS, keyframes=1, tolerance=1)
+@ngl.scene()
+def filter_selector_hue(_):
+    return _base_scene(
+        ngl.FilterSelector(
+            range=(-0.1 * math.tau, 0.45 * math.tau),
+            component="hue",
```
mbouron, line 156: Can you also exercise the "chroma" component ?

libnopegl/src/node_filters.c
```diff
@@ -191,6 +207,71 @@ static const struct node_param filtersaturation_params[] = {
 };
 #undef OFFSET
 
+enum {
+    SELECTOR_COMPONENT_LIGHTNESS,
+    SELECTOR_COMPONENT_CHROMA,
+    SELECTOR_COMPONENT_HUE,
```
mbouron, line 213: Can you explicitly set their value since `2` is directly referenced in the shader ?

mbouron (approved):
Except from my comment, LGTM.

tests/filter.py
ubitux: added

libnopegl/src/node_filters.c
ubitux: done for all 3 enums, with a comment to explain why

PR #226, opened 2024-02-10
Allow empty groups

jacereda:


pynopegl/init_header.py
```diff
@@ -195,8 +195,6 @@ def _args_setter(self, cython_setter, param_name, *args):
         return cython_setter(param_name, args)
 
     def _add_nodes(self, param_name, *nodes):
-        if hasattr(nodes[0], "__iter__"):
-            raise Exception(f"add_{param_name}() takes elements as positional arguments, not list")
```
ubitux, line 199: This is a bit tricky for several reasons:
- This check is here to prevent code like `group.add_children([a, b, c])` which is a common mistake
- removing this check indeed allows empty sized `nodes` array, but then in `_pynopegl.pyx` you'll get a `calloc(0, ...)`, which is implementation defined (could either return a pointer or a `NULL`), which ends up making `ngl.Group(children=[])` non deterministic (depending on the platform)
- The f64s handling should be consistent with that logic

So instead, may I suggest instead adding something along the lines of `if not nodes: return 0` (and same with `f64s` below)?

ubitux (changes requested):

pynopegl/init_header.py
jacereda: Right, changed it.

pynopegl/init_header.py
```diff
@@ -195,11 +195,15 @@ def _args_setter(self, cython_setter, param_name, *args):
...
 
     def _add_nodes(self, param_name, *nodes):
+        if not nodes:
+            return 0
         if hasattr(nodes[0], "__iter__"):
             raise Exception(f"add_{param_name}() takes elements as positional arguments, not list")
         return self._param_add_nodes(param_name, len(nodes), nodes)
 
     def _add_f64s(self, param_name, *f64s):
+        if not nodes:
```
ubitux, line 205: I believe it should be `if not f64s`

Also, can you update the commit message to something more generic like: `pynopegl: allow empty argument for list-based parameters`?

ubitux (changes requested):

pynopegl/init_header.py
jacereda: D'oh!

mbouron (approved):

PR #229, opened 2024-02-12
Compute viewport using the scene aspect ratio

ubitux:
Depends on https://github.com/NopeForge/nope.gl/pull/228

PR #230, opened 2024-02-12
gpu_ctx: remove viewport parameter from ngli_gpu_ctx_resize()

mbouron:
Depends on https://github.com/NopeForge/nope.gl/pull/229.


ubitux (approved):
Cool, should be fine after a rebase

PR #232, opened 2024-02-23
fontface: add index parameter

ubitux:
Tested with the Iosevka TTC font

### DIFF
--- a/libnopegl/src/node_fontface.c
+++ b/libnopegl/src/node_fontface.c
@@ -29,12 +29,19 @@ static const struct node_param fontface_params[] = {
     {"path",  NGLI_PARAM_TYPE_STR, OFFSET(path),
               .flags=NGLI_PARAM_FLAG_NON_NULL | NGLI_PARAM_FLAG_FILEPATH,
               .desc=NGLI_DOCSTRING("path to the font file")},
+    {"index", NGLI_PARAM_TYPE_I32, OFFSET(index),
+              .desc=NGLI_DOCSTRING("index of the face in the font file")},
     {NULL}
 };
 
 #if HAVE_TEXT_LIBRARIES
 static int fontface_init(struct ngl_node *node)
 {
+    const struct fontface_opts *o = node->opts;
+    if (o->index < 0 || o->index > 0xffff) {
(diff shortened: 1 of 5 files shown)


libnopegl/src/text_external.c
```diff
@@ -60,9 +60,9 @@ static int load_font(struct text *text, const char *font_file)
     FT_Face ft_face = NULL;
     hb_font_t *hb_font = NULL;
 
-    FT_Error ft_error = FT_New_Face(text->ctx->ft_library, font_file, 0, &ft_face);
+    FT_Error ft_error = FT_New_Face(text->ctx->ft_library, font_file, face_index, &ft_face);
```
mbouron, line 63: I think we should sanitize face_index against negative values, see: https://freetype.org/freetype2/docs/reference/ft2-face_creation.html#ft_open_face.

libnopegl/src/node_fontface.c
```diff
@@ -29,6 +29,8 @@ static const struct node_param fontface_params[] = {
     {"path",  NGLI_PARAM_TYPE_STR, OFFSET(path),
               .flags=NGLI_PARAM_FLAG_NON_NULL | NGLI_PARAM_FLAG_FILEPATH,
               .desc=NGLI_DOCSTRING("path to the font file")},
+    {"index", NGLI_PARAM_TYPE_I32, OFFSET(index),
+              .desc=NGLI_DOCSTRING("index of the face in the font file")},
```
mbouron, line 33: We should probably mention that the index is related to freetype `face_index` (https://freetype.org/freetype2/docs/reference/ft2-face_creation.html#ft_open_face).

libnopegl/src/text_external.c
ubitux: Added a `[0,0xffff]` range check at fontface init

libnopegl/src/node_fontface.c
ubitux: I'd rather not, freetype is an implementation detail. FT doc says "Bits 0-15 are the index of the face in the font file (starting with value 0)", so I believe using the description "index of the face in the font file" is consistent with that

mbouron (approved):

PR #245, opened 2024-03-23
Misc cleanups

ubitux:


pynopegl-utils/pynopegl_utils/tests/debug.py
```diff
@@ -26,6 +26,23 @@
 import pynopegl as ngl
 
 
+def get_grid_coordinates(cols: int, rows: int) -> Mapping[str, Tuple[float, float]]:
```
mbouron, line 29: Can we somehow align the names of the two functions ? `get_grid_points()` maybe ?

mbouron (approved):
Except from my comment, LGTM.

pynopegl-utils/pynopegl_utils/tests/debug.py
ubitux: They don't return the same type of data; one is a data dict and the other is a node tree; I'd rather not imply they're the same type of function

mbouron: But that is given by the functions signatures. `get_grid_coordinates()` is used like this `points=get_grid_coordinates(_N, _N)` so why not call it `get_grid_points()` ?

mbouron: I would also make sense to put that function in `cmp_cuepoints.py`.

ubitux: I have conflicting thoughts about all of this:
- "that is given by the functions signatures", yeah but well "get_{grid,debug}_points()" being totally different function types is misleading
- `get_debug_points()` might be misnamed, so maybe a change of name is required first
- having functions in `cmp_cuepoints.py` is weird because the `cmp_*` now only expose a single class/function, and it feels like hacking random utils into them; it would also imply moving `get_debug_points()` there because it's only used in the context of cuepoints

ubitux: Submitted a new proposition.

pynopegl-utils/pynopegl_utils/tests/cuepoints_utils.py
```diff
@@ -43,7 +43,7 @@ def get_grid_points(cols: int, rows: int) -> Mapping[str, Tuple[float, float]]:
     return points
 
 
-def get_debug_points(
+def get_point_nodes(
```
mbouron, line 46: `get_points_nodes()` ?

PR #246, opened 2024-03-27
scene,api: allow aspect ratio to be set to 0

mbouron:
Setting the aspect ratio to 0 allows the main viewport dimensions to match the output window/surface dimensions. This is particularly useful if one wants to render to the whole window/surface and do not want nope.gl to manage the aspect ratio.


ubitux (commented):
What happens in case of a resize?

mbouron:
> What happens in case of a resize?

The viewport is automatically computed to match the new window/surface dimensions.

ubitux (approved):
Should be OK, but I'd be more comfortable with a test.

Does it work correctly with node that use the aspect ratio?

mbouron:
> > What happens in case of a resize?
> 
> The viewport is automatically computed to match the new window/surface dimensions.

Good point. Added commits so the player and cmp module handle the aspect ratio being set to 0 plus a test that exercises text rendering in such scenario.

ubitux (approved):

PR #248, opened 2024-04-02
Bump external dep versions

mbouron:


configure.py
```diff
@@ -155,10 +155,10 @@ def _gen_android_cross_file(cfg):
         sha256="8e6efa8d9ec5ef1b2c735b2fc2afdfed210a3a5aa01c48e572326914e27bb221",
     ),
     sdl2_Windows=dict(
-        version="2.26.5",
+        version="2.30.1",
```
ubitux, line 158: Sorry for the late review; 2.30.2 was released 2 days ago

ubitux (approved):
Should be fine.

I'm seeing Unicode 15 support in Harfbuzz (https://github.com/harfbuzz/harfbuzz/blob/main/NEWS) in the interesting stuff. Did you notice anything in glslang (https://github.com/KhronosGroup/glslang/blob/main/CHANGES.md)?

configure.py
mbouron: Updated, thanks !

mbouron:
> Should be fine.
> 
> I'm seeing Unicode 15 support in Harfbuzz (https://github.com/harfbuzz/harfbuzz/blob/main/NEWS) in the interesting stuff. Did you notice anything in glslang (https://github.com/KhronosGroup/glslang/blob/main/CHANGES.md)?

Nothing major. `GL_EXT_control_flow_attributes2` support can be handy to ask for loop unrolling. `GL_ARB_texture_multisample` support is also nice if we make more use of MSAA in the future.

PR #250, opened 2024-04-06
Text cleanups

ubitux:


libnopegl/src/text.c
```diff
@@ -213,18 +213,21 @@ static void set_geometry_data(struct text *s, struct text_data_pointers ptrs)
...
         const float chr_corner_x = corner_x + width  * chr->geom.x;
         const float chr_corner_y = corner_y + height * chr->geom.y;
         const float transform[] = {chr_corner_x, chr_corner_y, chr_width, chr_height};
         memcpy(ptrs.pos_size + 4 * n, transform, sizeof(transform));
+    }
 
-        /* register atlas identifier */
+    /* register atlas identifier */
+    for (size_t n = 0; n < ngli_darray_count(&s->chars); n++) {
+        const struct char_info *chr = &chars[n];
```
mbouron, line 230: Just curious, did you measure any gain in practice ?

mbouron (approved):

mbouron:
```
 text: split loop to make memory access contiguous

This is similar to what's being done in fill_default_data_buffers():
ptrs.* contains locations in the same buffer, so this prevents random
accesses.
```
Did you measure any gain in practice ?

ubitux:
> Did you measure any gain in practice ?

No, nothing noticeable with my hardware, I'd probably have to use a large amount of data to spot this. It just makes the buffer reading completely linear since the it has the following layout: `[transform0][transform1][transform2]...[transformN][color0][color1][color2]...[colorN]...`. It's not too much change to split the loop so I'd rather do the sane thing here.

PR #251, opened 2024-04-09
Text transforms with anchors

ubitux:


libnopegl/src/text.c
```diff
@@ -308,14 +308,53 @@ static int set_vec3_value(float *dst, struct ngl_node *node, const float *value,
...
+    } else if (effect_opts->anchor_ref == NGLI_TEXT_ANCHOR_REF_VIEWPORT) {
+        anchor_x = effect_opts->anchor[0];
+        anchor_y = effect_opts->anchor[1];
+    } else {
+        ngli_assert(0);
+    }
+
+    NGLI_ALIGNED_MAT(tm);
+    NGLI_ALIGNED_MAT(treloc0);
+    NGLI_ALIGNED_MAT(treloc1);
```
mbouron, line 346: nit: make the names consistent (if that makes sense). I guess `tm` stands for `transformation matrix` while `treloc0` stands for transformation effect location 0 ?

mbouron (approved):
Other than my nit comment, LGTM.

libnopegl/src/text.c
ubitux: `tm` for "transform matrix" yes, `treloc` for "transform relocation"

What alternative do you suggest? `tmreloc*`?

mbouron: `tmreloc*` sounds good.

PR #252, opened 2024-04-10
nodes: add DrawMask node

ubitux:


libnopegl/src/node_drawother.c
```diff
@@ -399,6 +412,23 @@ static const struct node_param drawhistogram_params[] = {
 };
 #undef OFFSET
 
+#define OFFSET(x) offsetof(struct drawmask_opts, x)
+static const struct node_param drawmask_params[] = {
+    {"content",   NGLI_PARAM_TYPE_NODE, OFFSET(content),
+                 .flags=NGLI_PARAM_FLAG_NON_NULL,
+                 .node_types=(const uint32_t[]){TRANSFORM_TYPES_ARGS, NGL_NODE_TEXTURE2D, NGLI_NODE_NONE},
+                 .desc=NGLI_DOCSTRING("texture being masked")},
+    {"overlay",   NGLI_PARAM_TYPE_NODE, OFFSET(overlay),
```
mbouron, line 421: is there a reason why it is not called mask ?

libnopegl/src/node_drawother.c
```diff
@@ -399,6 +412,23 @@ static const struct node_param drawhistogram_params[] = {
 };
 #undef OFFSET
 
+#define OFFSET(x) offsetof(struct drawmask_opts, x)
+static const struct node_param drawmask_params[] = {
+    {"content",   NGLI_PARAM_TYPE_NODE, OFFSET(content),
+                 .flags=NGLI_PARAM_FLAG_NON_NULL,
+                 .node_types=(const uint32_t[]){TRANSFORM_TYPES_ARGS, NGL_NODE_TEXTURE2D, NGLI_NODE_NONE},
+                 .desc=NGLI_DOCSTRING("texture being masked")},
```
mbouron, line 420: if it's referenced as a texture in the doc, maybe change the field name to texture ? (or the other way around)

libnopegl/src/node_drawother.c
mbouron: Morever can you document that currently only the red channel is used for the masking operation ?

mbouron (approved):
Other than my remarks. LGTM.

libnopegl/src/node_drawother.c
ubitux: > is there a reason why it is not called mask ?

No particular reason, I guess just had something you apply on top in my mental model and wanted to avoid the redundancy between the node name and its parameter.

Renamed to `mask` and mentioned the red channel only thing.

PR #254, opened 2024-04-11
hblur: document that only the red channel of the blur map is used to provide the CoC values

mbouron:


libnopegl/src/node_hblur.c
```diff
@@ -107,7 +107,7 @@ static const struct node_param hblur_params[] = {
     {"map",         NGLI_PARAM_TYPE_NODE, OFFSET(map),
                     .node_types=(const uint32_t[]){NGL_NODE_TEXTURE2D, NGLI_NODE_NONE},
                     .flags=NGLI_PARAM_FLAG_DOT_DISPLAY_FIELDNAME,
-                    .desc=NGLI_DOCSTRING("blur map providing the CoC (circle of confusion) for each pixels")},
+                    .desc=NGLI_DOCSTRING("blur map providing the CoC (circle of confusion) for each pixels. Currently, only the red channel of the blur map is used to provide the CoC values")},
```
ubitux, line 110: I'd avoid making several sentences, we don't do that yet (it creates wonky rules with regards to capitalization and periods). How about the following compact/dense desc: "blur map providing the CoC (circle of confusion) for each pixels (only the red channel is used)"

mbouron: Updated.

PR #257, opened 2024-05-07
build: fix meson cpu families 

mbouron:
Depends on https://github.com/NopeForge/nope.gl/pull/256.


mbouron:
Superseeded by https://github.com/NopeForge/nope.gl/pull/258.

PR #269, opened 2025-05-11
Misc improvements

mbouron:


libnopegl/src/node_drawother.c
```diff
@@ -580,63 +579,59 @@ static int init(struct ngl_node *node,
...
-        s->position_attr.buffer = s->vertices;
+        ret = ngli_geometry_set_uvcoords(s->geometry, 4, default_uvcoords);
+        if (ret < 0)
+            return ret;
 
-        s->uvcoord_attr.stride = 2 * sizeof(float);
-        s->uvcoord_attr.buffer = s->uvcoords;
+        ret = ngli_geometry_init(s->geometry, NGLI_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP);
+        if (ret < 0)
+            return ret;
```
ubitux, line 597: Style: can we keep the existing compact pattern?

```c
        int ret;
        if ((ret = ngli_geometry_set_vertices(s->geometry, 4, default_vertices)) < 0 ||
            (ret = ngli_geometry_set_uvcoords(s->geometry, 4, default_uvcoords)) < 0 ||
            (ret = ngli_geometry_init(s->geometry, NGLI_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP) < 0))
            return ret;
```

libnopegl/src/node_draw.c
```diff
@@ -166,26 +168,80 @@ static const struct node_param render_params[] = {
     {NULL}
 };
 
-static int render_init(struct ngl_node *node)
+static int check_params(struct ngl_node *node)
```
ubitux, line 171: I think the node can be `const` if you're only checking parameters?

libnopegl/src/node_compute.c
```diff
@@ -114,10 +114,9 @@ static const struct node_param compute_params[] = {
     {NULL}
 };
 
-static int compute_init(struct ngl_node *node)
+static int check_params(struct ngl_node *node)
```
ubitux, line 117: ditto `const`

libnopegl/src/scene.c
```diff
@@ -150,6 +152,10 @@ static int setup_nodes(void *user_arg, struct ngl_node *parent, struct ngl_node
     if (parent) {
         if (!ngli_darray_push(&parent->children, &node))
             return NGL_ERROR_MEMORY;
+        if (node->cls->draw) {
+            if (!ngli_darray_push(&parent->draw_children, &node))
+                return NGL_ERROR_MEMORY;
+        }
```
ubitux, line 158: Why not merge the 2 conditions?

libnopegl/src/pass.h
```diff
@@ -30,6 +30,7 @@
 struct ngl_ctx;
 
 struct pass_params {
+    struct ngl_node *node;
```
ubitux, line 33: Is the pass allowed to change the node or can this be const?

libnopegl/src/node_rtt.c
```diff
@@ -79,6 +79,7 @@ static const struct node_param rtt_params[] = {
 };
 
 struct rtt_texture_info {
+    struct ngl_node *node;
```
ubitux, line 82: Can be const?

libnopegl/src/pass.c
```diff
@@ -174,21 +169,32 @@ static int register_texture(struct pass *s, const char *name, struct ngl_node *t
...
+                    type = NGLI_PGCRAFT_SHADER_TEX_TYPE_IMAGE_2D;
                 else if (texture->cls->id == NGL_NODE_TEXTURE2DARRAY)
-                    crafter_texture.type = NGLI_PGCRAFT_SHADER_TEX_TYPE_IMAGE_2D_ARRAY;
+                    type = NGLI_PGCRAFT_SHADER_TEX_TYPE_IMAGE_2D_ARRAY;
                 else if (texture->cls->id == NGL_NODE_TEXTURE3D)
-                    crafter_texture.type = NGLI_PGCRAFT_SHADER_TEX_TYPE_IMAGE_3D;
+                    type = NGLI_PGCRAFT_SHADER_TEX_TYPE_IMAGE_3D;
                 else if (texture->cls->id == NGL_NODE_TEXTURECUBE)
-                    crafter_texture.type = NGLI_PGCRAFT_SHADER_TEX_TYPE_IMAGE_CUBE;
+                    type = NGLI_PGCRAFT_SHADER_TEX_TYPE_IMAGE_CUBE;
```
ubitux, line 178: Already the case in the existing code but why is this almost exactly the same as the `type` assignment at the beginning of the function (except one is a switch case and the other is a bunch of if)?

libnopegl/src/scene.c
mbouron: Personal preference, I think it is easier to read.

libnopegl/src/node_drawother.c
mbouron: I'm not huge found of that style because the day you need to add some error logging you need to unwrap the whole pattern. For now, I can keep the style this way.

libnopegl/src/pass.c
mbouron: One is for texture/sampler types, the other one is for image (as in image load/store) types. I have some commit on the way that extract this to dedicated function. it makes the code cleaner (and the same style is used for both case).

libnopegl/src/node_draw.c
mbouron: Done !

libnopegl/src/pass.h
mbouron: Made it const.

libnopegl/src/node_rtt.c
mbouron: Const !

libnopegl/src/pass.h
ubitux: You did?

libnopegl/src/node_rtt.c
mbouron: Sorry I forgot to push. But, while it can be made const, we modify this priv data. So I think it is better to keep it that way.

ubitux (approved):

PR #270, opened 2025-05-11
Distmap fixes part1

ubitux:


libnopegl/src/glsl/distmap.frag
```diff
@@ -140,14 +140,23 @@ vec2 sum_of_inv(vec2 z0, vec2 z1, vec2 z2)                   { return c_inv(z0 -
 #define XERR 1e-6
 #define SMALL_OFF(off) (dot(off, off) <= XERR * XERR)
 
+float max_3(float a, float b, float c)                   { return max(a, max(b, c)); }
+float max_4(float a, float b, float c, float d)          { return max(a, max_3(b, c, d)); }
+float max_5(float a, float b, float c, float d, float e) { return max(a, max_4(b, c, d, e)); }
```
mbouron, line 145: I think yhou can also expose them as helper functions `ngli_max{3,4,5}()`.

mbouron (approved):

mbouron (approved):
Have you noticed/measured any performance improvement (not blocking).
Other than my remark, LGTM.

libnopegl/src/glsl/distmap.frag
ubitux: Removed the commit that drop the helper, and added these functions instead

ubitux:
> Have you noticed/measured any performance improvement

No I didn't measure, but:
1. on certain "tricky" bezier, the algorithm performs better (less glitches) because it's able to find roots in the 16 iterations where it couldn't before
2. if I lower the maximum number of iterations, the algorithm perform without glitching anymore on many scenes

We could almost consider halving the number of iteration at this point.

PR #271, opened 2025-05-11
gpu_ctx: add ngli_backend_get_{string_id,full_name}()

mbouron:


ubitux (commented):
Not a fan of that, there is already an API for that:
```python
>>> ngl.get_backends()[ngl.Backend.OPENGLES]["name"]
'OpenGL ES'
```
In C you make a call to `ngl_backends_get()` (there is no graphics initialization) and you can store the `ngl_backend **backendsp` as long as you need to access the string ID or names (`ngl_backend.{string_id,name}`)

mbouron:
> Not a fan of that, there is already an API for that:
> 
> ```python
> >>> ngl.get_backends()[ngl.Backend.OPENGLES]["name"]
> 'OpenGL ES'
> ```
> 
> In C you make a call to `ngl_backends_get()` (there is no graphics initialization) and you can store the `ngl_backend **backendsp` as long as you need to access the string ID or names (`ngl_backend.{string_id,name}`)

Dropped the public API part and moved those functions to gpu_ctx.

ubitux (approved):

PR #272, opened 2025-05-12
Split node private structures to dedicated headers

mbouron:
Based on https://github.com/NopeForge/nope.gl/pull/269.


libnopegl/meson.build
```diff
@@ -131,6 +131,7 @@ trim_prefix = run_command([
 ], check: true).stdout().strip()
 
 project_args = cc.get_supported_arguments([
+  '-Wshadow',
```
ubitux, line 134: how about the `ngl-tools`?

ubitux (approved):
Aside from my comment LGTM

PR #273, opened 2025-05-12
Add cubic bezier easing

mbouron:


libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+static float find_first_cubic_root(float p0, float p1, float p2, float p3)
+{
+    double a = 3.0 * (p0 - 2.0 * p1 + p2);
+    double b = 3.0 * (p1 - p0);
+    double c = p0;
+    double d = -p0 + 3.0 * (p1 - p2) + p3;
+
+    if (double_close_to(d, 0.0, FLT_EPSILON)) {
+        if (double_close_to(a, 0.0, FLT_EPSILON)) {
+            if (double_close_to(b, 0.0, FLT_EPSILON)) {
```
ubitux, line 370: If you want to prevent a division by zero you can check directly against 0.0, it is enough to cover all `NaN` cases (assuming the numerator is finite). So I'm assuming this check is to prevent an `Inf` value. Unfortunately you can't check for that like that. For example 3e-16 will pass the epsilon check, but 1e+308/3e-16 will make an `Inf`.

Note: you probably wanted to use `DBL_EPSILON` here since you're working with double precision. But in general, epsilon is never a good threshold against `NaN`/`Inf` (and actually using double epsilon here will make it worse).

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
 DECLARE_EASINGS(back,            , back_func(x, PARAM(0, 1.70158)), TRANSFORM)
 DECLARE_EASINGS(back, _derivative, back_derivative(x, PARAM(0, 1.70158)), DERIVATIVE)
 
+/* Cubic Bezier */
+
+static int double_close_to(double value, double ref, double eps)
+{
+    return fabs(value - ref) < eps;
```
ubitux, line 330: `ref` and `eps` are always the same in your calls, so it could probably be called `close_to_zero` and take only one parameter

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+}
+
+static float clamp_root(float r)
+{
+    const float s = NGLI_CLAMP(r, 0.f, 1.f);
+    return fabsf(s - r) > FLT_EPSILON ? NAN : s;
```
ubitux, line 336: `FLT_EPSILON` is, as far as I can tell, not a very good threshold choice as it's probably going to filter out all sorts of cases where the root is close to 0 or 1: it's a *very* restrictive value, and at the same time not 100%. A more arbitrary 1e-6 (that can be understood as a "precision" rounding) will likely do a better job.

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+ * Find the first cubic root using Cardano's algorithm.
+ *
+ * See: https://pomax.github.io/bezierinfo/#yforx
+ */
+static float find_first_cubic_root(float p0, float p1, float p2, float p3)
+{
+    double a = 3.0 * (p0 - 2.0 * p1 + p2);
+    double b = 3.0 * (p1 - p0);
+    double c = p0;
+    double d = -p0 + 3.0 * (p1 - p2) + p3;
```
ubitux, line 366: So I'm assuming this is the cubic points being converted to polynomial coefficients, but the value are weirdly shuffled. You made your polynomial dx³+ax²+bx+c, which I'm assuming you did because you're making it monic later down the line and reusing a/b/c naming. I think it would make sense to keep it as ax³+bx²+cx+d for clarity, then pass it down to another function as `f(b/a, c/a d/a)`

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+            return clamp_root((float) (-c / b));
+        } else {
+            double q = sqrt(b * b - 4.0 * a * c);
+            double a2 = 2.0 * a;
+
+            float root = clamp_root(((float) ((q - b) / a2)));
+            if (!isnan(root))
+                return root;
+
+            return clamp_root(((float) ((-b - q) / a2)));
```
ubitux, line 382: Non-blocking: there is a simpler quadratic formula which we're using in distmap where you find a middle point `m` then offset ±√Δ to get each root

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+    double d = -p0 + 3.0 * (p1 - p2) + p3;
+
+    if (double_close_to(d, 0.0, FLT_EPSILON)) {
+        if (double_close_to(a, 0.0, FLT_EPSILON)) {
+            if (double_close_to(b, 0.0, FLT_EPSILON)) {
+                return NAN;
+            }
+            return clamp_root((float) (-c / b));
+        } else {
+            double q = sqrt(b * b - 4.0 * a * c);
```
ubitux, line 375: This is missing a check if it's the sqrt parameter is < 0

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+
+        return clamp_root((float)(-u1 - a3));
+    }
+
+    // One real root, two complex roots
+    double sd = sqrt(discriminant);
+    double u1 = fast_cbrt((float)(-q2 + sd));
+    double v1 = fast_cbrt((float)(q2 + sd));
+
+    return clamp_root((float)(u1 - v1 - a3));
```
ubitux, line 431: Note that in my experience the analytic cubic is filled with numerical instability and I wouldn't recommend it... I ended up dropping it from the distmap for a more stable iterative algorithm

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+}
+
+static double evaluate_cubic(double p1, double p2, double t)
+{
+    double a = 1.0 / 3.0 + (p1 - p2);
+    double b = (p2 - 2.0 * p1);
+    double c = p1;
+    return 3.0 * ((a * t + b) * t + c) * t;
```
ubitux, line 439: I would recommend computing and caching the polynomial for faster evaluation

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+
+static easing_type bezier_cubic(easing_type t, size_t args_nb, const easing_type *args)
+{
+    if (t < 0.f || t > 1.f)
+        return t;
+
+    const easing_type a = PARAM(0, 0.0);
+    const easing_type b = PARAM(1, 0.0);
+    const easing_type c = PARAM(2, 1.0);
+    const easing_type d = PARAM(3, 1.0);
```
ubitux, line 450: So here is something: with easing, the starting and ending points are always respectively `(0,0)` and `(1,1)`. If you need some sort of offsetting, you can use `easing_start_offset` and `easing_end_offset`.

Next, you can't have the curve going backward, which means $b_x ≤ c_x$ must be verified.

All these constraints may also make the solver and friends way more convenient and may avoid a bunch of tests. You also won't need to extend the number of parameter from 2 to 4.

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+
+    const easing_type v = NGLI_MAX(t, FLT_EPSILON);
```
ubitux, line 452: why is t=0 not allowed? It should just be the starting point

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+    const float dp1 = (float)( da - v);
+    const float dp2 = (float)( dc - v);
+
+    easing_type r = find_first_cubic_root(dp0, dp1, dp2, 0.f);
+    if (isnan(r)) {
+        /* No root found */
+        return 1.0;
+    }
+
+    return evaluate_cubic(db, 0.f, r);
```
ubitux, line 492: `evaluate_cubic()` takes `double` arguments so `0.f` should be `0.0`.

libnopegl/src/node_animkeyframe.c
```diff
@@ -319,6 +323,175 @@ static easing_type back_derivative(easing_type t, easing_type s)
...
+        return 1.0;
+
+    const easing_type a = PARAM(0, 0.0);
+    const easing_type b = PARAM(1, 0.0);
+    const easing_type c = PARAM(2, 1.0);
+    const easing_type d = PARAM(3, 1.0);
+
+    const double da = 3.0f * (b - a);
+    const double db = 3.0f * (c - b);
+    const double dc = 3.0f * (d - c);
```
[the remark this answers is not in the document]
mbouron: Fixed.

libnopegl/src/node_animkeyframe.c
mbouron: Done.

libnopegl/src/node_animkeyframe.c
mbouron: I implemented it so it matches the ref mentioned in the comment, see: https://pomax.github.io/bezierinfo/#yforx

libnopegl/src/node_animkeyframe.c
mbouron: Fixed.

PR #290, opened 2025-06-17
Simplify ngpu program interface (gl)

mbouron:
```
Shader probing is used exclusively in program_gl_utils to set the locations and
bindings of the program for older versions of OpenGL that do not support
explicit locations and bindings. Additionally, probing the attribute locations
is solely used for determining whether the shader requires re-linking, a
process that can be inlined directly into program_gl_utils.
```

Simplifies the maintenance of the GL backend !


ubitux (approved):
LGTM but you might want to go further (see my comment)

PR #292, opened 2025-06-17
Misc cleanups

ubitux:


libnopegl/src/api.c
```diff
@@ -55,10 +55,10 @@
...
 
 // Ensure public enums are stored on 32-bit
-NGLI_STATIC_ASSERT("enum_on_32bit", sizeof(enum ngl_log_level) == sizeof(int32_t));
-NGLI_STATIC_ASSERT("enum_on_32bit", sizeof(enum ngl_platform_type) == sizeof(int32_t));
-NGLI_STATIC_ASSERT("enum_on_32bit", sizeof(enum ngl_backend_type) == sizeof(int32_t));
-NGLI_STATIC_ASSERT("enum_on_32bit", sizeof(enum ngl_capture_buffer_type) == sizeof(int32_t));
+NGLI_STATIC_ASSERT("32-bit loglevel enum", sizeof(enum ngl_log_level) == sizeof(int32_t));
+NGLI_STATIC_ASSERT("32-bit platform enum", sizeof(enum ngl_platform_type) == sizeof(int32_t));
+NGLI_STATIC_ASSERT("32-bit backend enum",  sizeof(enum ngl_backend_type) == sizeof(int32_t));
+NGLI_STATIC_ASSERT("32-bit capture enum",  sizeof(enum ngl_capture_buffer_type) == sizeof(int32_t));
```
mbouron, line 61: Question, would you agree to change the API from ASSERT("string", cond) to ASSERT(cond, "string") so it matches _Static_assert() / static_assert() ?
(I'm not asking you to do it, I can handle that if you agree).

ubitux: Sure, no issue with that

PR #297, opened 2025-06-21
shaders: report line numbers error accurately

ubitux:


libnopegl/src/ngpu/opengl/program_gl.c
```diff
@@ -120,12 +118,7 @@ int ngpu_program_gl_init(struct ngpu_program *s, const struct ngpu_program_param
         gl->funcs.CompileShader(shader);
         ret = program_check_status(gl, shader, GL_COMPILE_STATUS);
         if (ret < 0) {
-            char *s_with_numbers = ngli_numbered_lines(shaders[i].src);
-            if (s_with_numbers) {
-                LOG(ERROR, "failed to compile shader \"%s\":\n%s",
-                    params->label ? params->label : "", s_with_numbers);
-                ngli_free(s_with_numbers);
-            }
+            LOG(ERROR, "failed to compile shader \"%s\"", params->label ? params->label : "");
```
mbouron, line 121: I tested a bit the PR, and for now, I think it makes debugging harder. While it works for simple cases, it does not scale well, especially since we are injecting the ressources / inputs / outputs / helpers, ie: there's no other way to visualize the whole assembled shader, which is of great help debugging. This is even more true for builtin shaders assembled from different glsl helpers and/or when `nopegl` is distributed across multiple devices with different kind of gpus (for example, one shader can fail on a specific hw / driver combinaison and works nicely on the other devices). Just having the error reported without the full context would make debugging a nightmare.

ubitux: With a debug option we could add a flag to keep dumping the whole thing for the project developers.

For users though the current experience as-is is much better experience. They don't have access to includes so and the line reset appears right before their input, which means the lines are aligned with their input, always (unless they have their own include system, in which case they can also use line directives in their own preprocessor).

PR #304, opened 2025-06-29
Misc cleanups

ubitux:


libnopegl/meson.build
```diff
@@ -586,7 +586,7 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
         {'name': 'OGLCompiler',                     'required': false},
         {'name': 'GenericCodeGen',                  'required': false},
         {'name': 'SPVRemapper',                     'required': true},
-        {'name': 'SPIRV',                           'required': true},
+        {'name': 'SPIRV',                           'required': false},
```
mbouron, line 589: Can you raise the min version requirement of glslang to `15.3.0` (it's in the same file) ?

ubitux: I don't mind but if we do that it means we can drastically simplify this code:
- `HLSL` and `OGLCompiler` don't exist anymore
- `MachineIndependent`, `OSDependent`, `GenericCodeGen` and `SPIRV` are now stubs so that can be dropped as well (they wouldn't be needed in a static link even if present)
- the comment could probably go away

Would you like me to do this?

mbouron (approved):
Other than my remarks, LGTM.
Thanks a lot, especially for moving python to a dedicated sub directory (and fixing the pyright integration) !
GG

libnopegl/meson.build
ubitux: I added a commit to do the bump, but not squashed with this one in case we want to revert it. This particular commit is here to allow SPIRV not to be present, so that when they actually drop it entirely our build would still work for new versions.

PR #307, opened 2025-06-30
Make dimensions unsigned and enable -Wsign-conversion

mbouron:


libnopegl/src/hud.c
```diff
@@ -531,13 +531,13 @@ static void draw_block_graph(struct hud *s,
 {
     const int64_t graph_h = graph_max - graph_min;
     const float vscale = (float)rect->h / (float)graph_h;
-    const int start = (d->pos - d->count + d->nb_values) % d->nb_values;
+    const size_t start = (d->pos - d->count + d->nb_values) % d->nb_values;
```
ubitux, line 534: `count` is always larger or equal to `pos`, so the subtraction goes into unsigned underflow. The arithmetic still works out but do we want this or should we for example reorder the operation to put the subtract at the end to avoid the underflow?

libnopegl/src/hud.c
```diff
@@ -661,7 +661,9 @@ static void widget_memory_draw(struct hud *s, struct widget *widget)
         else
             snprintf(buf, sizeof(buf), "%-12s %zuG", label, size / (1024 * 1024 * 1024));
         print_text(s, widget->text_x, widget->text_y + (int)i * NGLI_FONT_H, buf, color);
-        register_graph_value(&widget->data_graph[i], size);
+
+        const int64_t size_i64 = size >= INT64_MAX ? INT64_MAX : (int64_t)size;
```
ubitux, line 665: `NGLI_MIN(size, INT64_MAX)`?

libnopegl/src/distmap.c
```diff
@@ -386,8 +386,8 @@ static int draw_glyphs(struct distmap *s)
     if (ret < 0)
         return ret;
 
-    ngli_pipeline_compat_update_buffer(s->pipeline_compat, 0, s->vert_buffer, 0, (int)s->vert_offset);
-    ngli_pipeline_compat_update_buffer(s->pipeline_compat, 1, s->frag_buffer, 0, (int)s->frag_offset);
+    ngli_pipeline_compat_update_buffer(s->pipeline_compat, 0, s->vert_buffer, 0, (uint32_t)s->vert_offset);
+    ngli_pipeline_compat_update_buffer(s->pipeline_compat, 1, s->frag_buffer, 0, (uint32_t)s->frag_offset);
```
ubitux, line 390: isn't this argument a `size_t`?

libnopegl/src/distmap.c
```diff
@@ -474,7 +474,7 @@ int ngli_distmap_finalize(struct distmap *s)
         return NGL_ERROR_INVALID_USAGE;
     }
 
-    const int32_t nb_shapes = (int32_t)ngli_darray_count(&s->shapes);
+    const size_t nb_shapes = ngli_darray_count(&s->shapes);
```
ubitux, line 477: I feel like we should probably add a guard in `ngli_distmap_add_shape()` to refuse adding more than `INT32_MAX` shapes

libnopegl/src/backends/gl/hwmap_videotoolbox_darwin_gl.c
```diff
@@ -98,9 +98,9 @@ static int vt_darwin_map_plane(struct hwmap *hwmap, IOSurfaceRef surface, size_t
 
     size_t width = IOSurfaceGetWidthOfPlane(surface, index);
     size_t height = IOSurfaceGetHeightOfPlane(surface, index);
-    if (width > INT_MAX || height > INT_MAX)
+    if (width > UINT32_MAX || height > UINT32_MAX)
```
ubitux, line 101: do we really support up to this range? I saw some GL calls in this diff that seem to only accept signed ints. Or in cases such as the mipmap thing.

libnopegl/src/ngpu/vulkan/texture_vk.c
```diff
@@ -676,9 +676,9 @@ static VkResult texture_vk_upload(struct ngpu_texture *s, const uint8_t *data, c
                 .layerCount     = 1,
             },
             .imageOffset = {
-                transfer_params->x,
-                transfer_params->y,
-                transfer_params->z,
+                (int32_t)transfer_params->x,
+                (int32_t)transfer_params->y,
+                (int32_t)transfer_params->z,
```
ubitux, line 681: I think it makes sense to keep those signed, they are offset that could be negative; vulkan kept them signed on purpose as far as I can tell

CHANGELOG.md
```diff
@@ -61,6 +61,9 @@ Versioning](https://semver.org/spec/v2.0.0.html) for `libnopegl`.
   opened
 - Text outline is now by default on the outer edge, and thus doesn't affect the
   shape of the characters anymore
+- Texture2D.{width,height}, Texture2DArray.{width,height,depth},
+  Texture3d.{width,height,depth}, TextureCube.size, TextureView.layer are now
+  unsigned
```
ubitux, line 66: This commit includes changes in the public API header which aren't documented (and probably belong in a separate commit)

libnopegl/src/ngpu/vulkan/texture_vk.c
mbouron: WGPU, SDL3 have them unsigned.

mbouron: Also Vulkan will error out if they are negative.

libnopegl/src/backends/gl/hwmap_videotoolbox_darwin_gl.c
mbouron: GL mostly use signed only, however, the modern api use  unsigned for dimensions. This check is only there because ngpu dimensions are now unsigned.

libnopegl/src/distmap.c
mbouron: Fixed.

libnopegl/src/hud.c
mbouron: Added with a cast.

libnopegl/src/hud.c
ubitux: Let's give `pos=3`, `count=10`, and `nb_values=10`.

In unsigned space:
- `(pos-count+nb_values)%nb_values = (3-10+10)%10 = 3`
- `(pos-count)%nb_values = (3-10)%10 = 18446744073709551609%10 = 9`

So nah.

libnopegl/src/hud.c
```diff
@@ -549,17 +549,17 @@ static void draw_line_graph(struct hud *s,
 {
     const int64_t graph_h = graph_max - graph_min;
     const float vscale = (float)rect->h / (float)graph_h;
-    const int start = (d->pos - d->count + d->nb_values) % d->nb_values;
+    const size_t start = (d->pos - d->count + d->nb_values) % d->nb_values;
```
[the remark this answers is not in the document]
mbouron: Added a separate commit to re-order the operations.

CHANGELOG.md
ubitux: Thanks.

> [!NOTE]
> There is no need to segment so much, `Texture2D.{width,height}` as a whole is just fine

PR #315, opened 2025-07-15
ngpu/glcontext: enable sRGB framebuffer conversions

mbouron:


ubitux (approved):
Probably OK, tho with srgb=False the test fails with only 2% error. This is probably because the fingerprint works on local differentiation; a color picker would probably be more reliable if you're comparing sRGB vs linear colors.

mbouron:
> Probably OK, tho with srgb=False the test fails with only 2% error. This is probably because the fingerprint works on local differentiation; a color picker would probably be more reliable if you're comparing sRGB vs linear colors.

I agree. I did that because of laziness. I changed the test.

PR #317, opened 2025-08-05
hwmap: access frame information prior to map_frame() call

mbouron:
The map_frame() callback may take ownership of the frame, depending on the interop implementation. This means the frame could be released at any point, making the frame pointer potentially invalid after the map_frame() call.

This commit notably fixes HDR playback on Android (Mediacodec) where the transfer function was accessed on a frame that had already been released.


ubitux (approved):

ubitux:
BTW, I noticed `frame->ts` is later down the line as well; not an issue?

